PR #41, opened 2020-06-17
Check to see if the user has WCS within their imaging data. If not, w…

tamimfatahi:
…ill plate solve for them to get WCS info.

### DIFF
--- a/exotic.py
+++ b/exotic.py
@@ -99,6 +99,7 @@ from photutils import aperture_photometry
 import astropy.units as u
 from astropy.time import Time
 from astropy.coordinates import SkyCoord, EarthLocation, AltAz
+from astropy.wcs import WCS
 import astroalign as aa
 
 # cross corrolation imports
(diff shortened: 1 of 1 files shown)


exotic.py
```diff
@@ -457,6 +458,61 @@ def getAirMass(hdul, raStr, decStr, lati, longit, elevation):
...
 
+# Check for WCS in the user's imaging data and possibly plate solve.
+def check_wcs(fits_files, saveDirectory):
+
+    # Opens the first input file and checks to see if it has WCS in the header
+    hdulist = fits.open(fits_files[0])
+    header = hdulist[0].header
+
+    # MJD seems to be throwing off an error. I deleted it since not important for plate solving
+    del header['MJD-OBS']
```
rzellem, line 469: Does this just delete the key ['MJD-OBS'] from the header variable? I.e., is this value still accessible from the fits file itself? We use that key later on when loading the time of the image.

tamimfatahi: This fits file is created only for the purpose for the user to check their x,y coordinates matching with ra and dec in ds9. Also, this fits file will be used to double-check the inputted values from the user. This will not affect the directory in which their imaging files are stored. It just uses that first imaging file to create a plate solution w/ WCS titled 'newfits.fits', similar to the astrometry one where the program doesn't use.

PR #42, opened 2020-06-19
New nasa API + code clean up

pearsonkyle:


exotic.py
```diff
@@ -25,7 +25,7 @@
 # Major releases are the first digit
 # The next two digits are minor commits
 # (If your commit will be #50, then you would type in 0.5.0; next commit would be 0.5.1)
-versionid = "0.7.5"
```
rzellem, line 28: Let's not increment the versionid here - let's do that when we merge dev into main. Also, I like to incrementally update, so the next versionid would be 0.7.6. Please remove your edit on this line for now.

exotic.py
```diff
@@ -1508,57 +1347,9 @@ def realTimeReduce(i):
...
-            # decMin = noSpaceColonDec[3:5]
-            # decSec = noSpaceColonDec[5:]
-            # if decSign == '-':
-            #     decDeg = round((float(decD) + float(decMin) / 60.0 + float(decSec) / 3600.0),
-            #                 4) * -1  # account for the negative
-            # else:
-            #     decDeg = round((float(decD) + float(decMin) / 60.0 + float(decSec) / 3600.0), 4)
-            decDeg = astropy.coordinates.Angle(decStr+" degrees").deg
+            # get coordinates from NASA exoplanet archive
+            raDeg = pDict['ra']
```
rzellem, line 1351: Does this work if the target is a TOI? Some of our participants are in TFOP and want to use EXOTIC to reduce their data. These targets might *not* be in SIMBAD or the Archive. Need to make sure this will still work for them.

jpl-jengelke:
Interleaving the code cleanup means there may be difficulty to merge back changes from other branches being worked, especially if they did their own rearranging or cleanup. As an FYI it may be prudent to check my assertion by attempting a merge from another issue current issue branch into a **copy** of this branch.

exotic.py
```diff
@@ -1735,8 +1526,10 @@ def realTimeReduce(i):
         print('*******************************************')
         print("Planetary Parameters for Lightcurve Fitting")
         print('')
+        
+        # based on the logic above this will never be false...
+        if not CandidatePlanetBool:
```
rzellem, line 1531: It will be false if the planet cannot be found in the NASA Exoplanet Archive (i.e., a TESS TOI).

jpl-jengelke:
Also, did you want to implement some retry logic? It's flexible enough to support short retry bursts over 10 to 20 second intervals. If there is a momentary network glitch or a transient outage, this would save the app from potentially having a bad execution or needing a restart. Because it's a consumer product I would recommend shortening the timeout for Requests connections to something like 10 secs. and bursting retries three times every 15 secs. on exception detection.

exotic.py
jpl-jengelke: Given the scope of the changes, the [Semantic Versioning Specification](https://semver.org/) implies that the [minor version](https://semver.org/#spec-item-7) (e.g. **Major.Minor.Patch**) should be updated given that the API ingestion change is a fairly different and new piece of functionality. But this is not a steadfast rule, of course, just a suggestion.

exotic.py
pearsonkyle: I've added this back in for the planet candidates. The TOIs will likely need their parameters added manually

exotic.py
pearsonkyle: Added back in, that comment was from an old iteration.

pearsonkyle:
> Also, did you want to implement some retry logic? It's flexible enough to support short retry bursts over 10 to 20 second intervals. If there is a momentary network glitch or a transient outage, this would save the app from potentially having a bad execution or needing a restart. Because it's a consumer product I would recommend shortening the timeout for Requests connections to something like 15 to 20 secs. and bursting retries three times using something like Tenacity every 15 secs. on exception detection.

A retry would be nice but depending on the user's internet speed the request may take up to 2-3 minutes. I had issues over the weekend with a 90 second timeout killing the request before it returned the data.

exotic.py
pearsonkyle: I'll let whoever merges with main to handle this

rzellem:
That’s you and me. ☺

PR #59, opened 2020-06-23
Update version to v0.8.3: Adding more detailed instructions and typo bug fix.

rzellem:


jpl-jengelke:
c8af419 Observations on "Instructions-How-to-Use-EXOTIC.pdf": 
- Name of latest contributor is missing from top right corner. I think it should be added. 
- Under "Changing Directories"
  - On Windows `cd` prints the current working directory (equivalent to `pwd` on Linux). This could be made clearer. 
  - "Command Prompt" is used to describe commands initially while "Native Windows" is used near the end. I would make it uniform and simplify it by stating "Windows CMD Terminal" 
  - To get to the home directory on Windows CMD Terminal, use `cd ~`
- Under "Listing Files in a Directory":
  - To perform a directory listing in Windows CMD Terminal, use `dir`
- "Quitting the Application" could be perhaps renamed more appropriately as "Quitting the Terminal":
  - The `exit` command works on both Windows and Shell. (Closing with the 'x' can potentially leave processes running in the background.)
- Under "Getting the Code to Run":
  - the `python3 ...` command is not always available on Windows. I would suggest the user try `python ...` if `python3` yields an error.

jpl-jengelke:
Comments on the recommended installation script for Linux, e.g. [exotic_installation_linux.sh](https://github.com/rzellem/EXOTIC/blob/develop/exotic_installation_linux.sh): 
- Consider switching out of the Downloads directory which is routinely ephemeral. Just go to root with `cd ~`.
- To prevent overwrite of a directory already named `EXOTIC` and to make sure it's current, something like:
  - `if [[ -d "EXOTIC/.git" ]] ; then  cd EXOTIC && git pull --all --prune --tags ; elif [[ -d "EXOTIC" ]] ; then mv EXOTIC EXOTIC-bak ; git clone ... ; cd EXOTIC ; else git clone ... ; cd EXOTIC ; fi` 
- I would move the `sudo apt-get -y install g++` to the top with other `apt-get` commands. 
  - Maybe consider an alternative `sudo apt-get -y install build-tools` which I think installs multiple components, including GCC/G++
- Then, perform any requisite installs first, ala `pip install --upgrade pip` and `pip install setup-tools`
- Then, most importantly, I would `cd` into the EXOTIC directory and execute `pip install -r requirements.txt` This will allow the installation directions to stand alone while application dependencies evolve as development progresses. 
- Then, simply execute `chmod 755 exotic.py && python3 exotic.py || python exotic.py`

jpl-jengelke:
Comments on the recommended installation script for Mac, e.g. [exotic_installation_mac.sh](https://github.com/rzellem/EXOTIC/blob/develop/exotic_installation_mac.sh): 
- Similar to the Linux installation ... 
  - Consider switching out of the Downloads directory which is routinely ephemeral. Just go to root with `cd ~` or to `cd ~/Documents` so it stands out to Mac users.
  - Same thing to prevent overwrite of a directory already named `EXOTIC` ala the scripting conditional.
  - Same with the requirements and `pip install -r ...`
  - Same thing with execution to set the file perms and fall back to `python` if `python3` fails. (It's okay. ... If Python 3 isn't really there users should see an error and it will stop in Python 2. ... A version test can always be added to the top of the application when it launches to actually verify it is in Python 3+.)

jpl-jengelke:
This is how to check in the application if Python is running with version 3: 
```
import sys
vs_list = sys.version.split('.')
print(vs_list)
print(vs_list[0] == '3')
# of course the MINOR version can also be validated if desired, e.g. 
minor = int(vs_list[1]) if vs_list[1].isdigit() else 0
print(minor >= 7)
```
alternately, 
```
import platform
print (platform.python_version())
# ... similar tests ... 
```

Issue #15, opened 2020-06-12
Fix limb darkening scraper to match user-input filter


rzellem:
Currently we scrape for the limb darkening parameters for the V-band, but some users will be observing in other bands. Need to scrape these based on the user's input.

tamimfatahi:
Completed in Pull Requests #60

PR #60, opened 2020-06-24
Bugs/Formatting 

tamimfatahi:
Filter names are now optional and will inform users if the wrong one is entered with a link to AAVSO for options of different ones. Also, Time may have caused some user's issues and is now being referenced directly through astropy.time. Minor formatting changes as well.


tamimfatahi:
Forgot to add, this ties in with issue #15 .

Issue #75, opened 2020-06-25
More robust limb darkening coefficient calculations


rzellem:
Right now, if a user enters an AAVSO-support limb darkening coefficient that is NOT listed on the EXOFAST website (http://astroutils.astronomy.ohio-state.edu/exofast/limbdark.shtml), then an error will be raised.

Perhaps we allow the user to enter an AAVSO-supported LD coefficient. But if it is not on the EXOFAST website, we allow them to choose the appropriate one from the list...?

pearsonkyle:
I'd say pick the closest filter from the set that is available to EXOFAST (U,B,V,R,I,J,H,K) 
If you want to get really fancy, LDTK allows one to calculate coefficients for an arbitrary filter if you know the wavelength region: https://github.com/hpparvi/ldtk

rzellem:
Yeah we should make the jump to LDTK and calculate this ourselves. Want to assign yourself to that task and work with Tamim, Kyle?

Sent from my iPhone

On Jun 25, 2020, at 6:25 PM, Kyle A. Pearson <notifications@github.com> wrote:

﻿

I'd say pick the closest filter from the set that is available to EXOFAST (U,B,V,R,I,J,H,K)
If you want to get really fancy, LDTK allows one to calculate coefficients for an arbitrary filter if you know the wavelength region: https://github.com/hpparvi/ldtk

tamimfatahi:
Completed in pull request #80

Issue #74, opened 2020-06-25
More robust planet name filtering


rzellem:
Right now, users are getting an error if they input "HAT-P-32b" instead of having a space between the host star name and the planet letter. If they input the planet name in this manner, an error is raised. Maybe have EXOTIC correct the planet name to add the space and then have EXOTIC ask the user to check the changed name.

For example:

User inputs "HAT-P-32b"
EXOTIC sees the lack of a space and says to the user:
"Did you mean HAT-P-32 b? (y/n)" or something similar
Then the user can confirm.

pearsonkyle:
Here are some planet names to stress test: 

HD 189733 b
GJ 1214 b
CoRoT-1 b

I added a feature to the code to be case insensitive. I don't think there will ever be a planet that requires case to distinguish it self

jpl-jengelke:
The planet identifier can be split off using the host star name. Then a space can be reintroduced. The real problems arise when spaces and dashes are moved about.

tamimfatahi:
Completed in pull request #80

Issue #16, opened 2020-06-12
Add EXOTIC's own limb darkening calculation code


rzellem:
Currently we scrape the EXOFAST limb darkening parameters; need to explore adding LDTK into EXOTIC to calculate it ourselves.

tamimfatahi:
Completed in pull request #175

PR #80, opened 2020-06-26
Planet Search and Pixel Scale

tamimfatahi:
Both the planet search from issue #74 and pixel scale are now more robust. Also, changed the site for the user to look up filter names by routing them to EXOFAST rather AAVSO for now in issue #75 . I say we move issue #75 to close and start looking towards issue #16 to solve the limb darkening calculations as I am looking into it now.

### DIFF
--- a/exotic.py
+++ b/exotic.py
@@ -52,6 +52,7 @@ def animate():
     # sys.stdout.write('\nDone!     \n')
     # print('\nDone!     \n')
 
+
 done = False
 t = threading.Thread(target=animate, daemon=True)
 t.start()
@@ -198,6 +199,7 @@ def dataframe_to_jsonfile(dataframe, filename):
     with open(filename, "w") as f:
         f.write(json.dumps(jsondata['data'], indent=4))
 
+
 def tap_query(base_url, query, dataframe=True):
     # table access protocol query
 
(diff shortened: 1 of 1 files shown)


exotic.py
```diff
@@ -344,16 +349,16 @@ def getJulianTime(hdul):
             exptime_offset = hdul[0].header['EXPTIME'] / 2. / 60. / 60. / 24.  # assume exptime is in seconds for now
 
     # If the mid-exposure time is given in the fits header, then no offset is needed to calculate the mid-exposure time
-    return (julianTime + exptime_offset)
+    return julianTime + exptime_offset
```
rzellem, line 352: Huh...I thought the () were needed for a return function. Guess not! TIL.

exotic.py
```diff
@@ -1646,7 +1656,8 @@ def parse_args():
                     quadLimb = float(quadString)
                     break
                 except:
-                    filterName = input('\nNot valid filter name. Please enter a valid filter name using aavso.org/filters: ')
+                    filterName = input('\nNot valid filter name. Please enter a valid filter name using '
```
rzellem, line 1659: Ultimately I think pointing them to the EXOFAST website is correct, but we will need to keep this in mind when we ingest data from this version of EXOTIC into CITISENS. We should think about contacting Jason Eastman at EXOFAST to see exactly what wavelengths/passbands correspond to each of the listed filters on EXOFAST.

PR #86, opened 2020-06-29
Issue 19

FuwaFuwaKamiSama:
Variable Star Check was implemented, testing and debugging on other devices would be greatly appreciated.


exotic.py
```diff
@@ -1717,6 +1709,35 @@ def parse_args():
...
+
+                # Plate solves the .fits file and returns status
+                WCS_fits_file = plate_solution(first_fits, saveDirectory)
+
+                # Get WCS for the first .fits file to track x,y coordinates
+                # WCS_fits_data = WCS_fits_file.getdata(WCS_fits_file)
+                done = True
+
+            #Doublechecking to see if comparison star != variable star
+            print("Verifying selected comparison star is not variable. Please wait.")
```
rzellem, line 1734: I don't think this is the right place for the comp star checker as this is not in the for loop that loops over the multiple comparison stars. Shouldn't this be after line that has `for compCounter in range(0, len(compStarList)):` instead? That way if the comparison star chosen is variable, the for loop skips over it. Or perhaps

exotic.py
```diff
@@ -523,6 +540,25 @@ def plate_solution(fits_file, saveDirectory):
...
+#Checks if comparison star coordinates don't point to variable stars
+def variableStarCheck(comparisonStarFile):
+    #Open WCS file and convert comparison star's coordinates into normal coordinates
+    starList = fits.open(comparisonStarFile)
+    wcsCoords = WCS.WCS(starList[0].header)
+    starCoords = wcsCoords.all_pix2world()
+
+    #Use normalized Coordinates to search on SIMBAD to check if that star is variable
+    try:
+        resultTable = Simbad.query_region(coord.SkyCoord(starCoords[0], starCoords[1], unit = (u.deg, u.deg))
```
rzellem, line 552: I don't see you looping over different stars in the image or passing the RA+DEC (or pixel location) of the comparison star. Looks like you pass the fits file to your function, open the fits file, load in the WCS coordinates, convert the pixel values to RAs and DECs, and then read in `starCoords[0]` and `starCoords[1]`, which are all (?) the RAs and DECs in the file. Or am I missing something?

exotic.py
```diff
@@ -523,6 +540,25 @@ def plate_solution(fits_file, saveDirectory):
...
+    except TypeError:
+        print("Error: Invalid star coordinates. Aborting process")
```
rzellem, line 554: If the user gets to this line of the code (i.e., they get a `TypeError`, resultTable variable will not be initialized, and they will get an error at line 557. Suggest just putting in a `return False` line here after this print statement or do something else.

exotic.py
```diff
@@ -541,23 +541,36 @@ def plate_solution(fits_file, saveDirectory):
...
+            validCoordList.append((compStarCoords[0], compStarCoords[1]))
 
+    #Use normalized Coordinates to search on SIMBAD to check if that star is variable
+    #Throws an error if invalid coordinates are inputted
+    for counter in range(0, len(validCoordList)):
+        try:
+            resultTable = Simbad.query_region(coord.SkyCoord(validCoordList[counter][0], validCoordList[counter][1], unit = (u.deg, u.deg)))
+        except TypeError:
+            print("Error: Invalid star coordinates. Checking next star... ")
+
```
rzellem, line 566: I think you're missing a `continue` here- otherwise, the rest of the `for` loop will continue to run and throw you an error.

exotic.py
```diff
@@ -2246,6 +2252,14 @@ def lc2min(x):
                     # Exit aperture loop
                 # Exit annulus loop
             # Exit the Comp Stars Loop
+            #Doublechecking to see if comparison star != variable star
```
rzellem, line 2255: I would move this up to the top of the `for` loop- that way, if a comp star is variable, you throw it away immediately- and don't waste time doing processes such as extracting its photometry - and move on right away to the next comp star.

exotic.py
FuwaFuwaKamiSama: I moved this to be right after the part that determines the optimal apertur for the comparison star as suggested since by then we'll have a list of comparison stars to go through.

exotic.py
FuwaFuwaKamiSama: I just added that per your suggestion; I didn't realize that was a possibility the first time through

exotic.py
FuwaFuwaKamiSama: That makes sense since we could save time by not having to go through twice to see which stars are variable instead of throwing them out before they ever get the chance. I'll figure out a better place to move the command to in the loop.

exotic.py
```diff
@@ -587,6 +593,38 @@ def plate_solution(fits_file, saveDirectory):
             return False
         time.sleep(5)
 
+#Checks if comparison star coordinates don't point to variable stars
+def variableStarCheck(rafile, decfile, compStarList):
+    validCoordList = []
+    errorMargin = 0.005556
```
rzellem, line 599: What is this errorMargin coming from?

exotic.py
rzellem: Still not seeing the `continue` command here.

exotic.py
```diff
@@ -587,6 +593,38 @@ def plate_solution(fits_file, saveDirectory):
...
+            print("Error: Invalid star coordinates. Checking next star... ")
+
+        #Check if star ID/name explicitly has V* in it
+        starName = resultTable['MAIN_ID'][0].decode("utf-8")
+        if "V*" in starName:
+            print("Comparison star " + starName + " is variable; selecting alternative star...")
+            compPixelCoords = validCoordList[counter].all_world2pix()
+            #validCoordList.pop(validCoordList[counter])
+            compStarList.remove(compPixelCoords)
+    return True
```
rzellem, line 627: the way your function runs, it will *always* return `True` - how does it behave differently if the star is variable or not? Don't you want to return instead the list of stars that are *not* variable?

FuwaFuwaKamiSama:
> Please see requested changes. Please also pull from develop to be up to date with latest release. Lastly, you have some conflicts still lingering in your code (look for the block of `>>>>>>>`) which needs to be resolved.

Resolved

exotic.py
FuwaFuwaKamiSama: Please see updated code

exotic.py
FuwaFuwaKamiSama: Please see updated code

exotic.py
FuwaFuwaKamiSama: Please see updated code

FuwaFuwaKamiSama:
> Changes needed for new function and for placement of the function in main. See comments for details.

Please see updated code

exotic.py
FuwaFuwaKamiSama: Added to updated version of code not yet pushed (both of us were present to see me add this to the code)

rzellem:
Closing for now while @FuwaFuwaKamiSama continues coding.

Issue #34, opened 2020-06-16
Allow users to ID non-standard image extensions


rzellem:
E.g., .fts

tamimfatahi:
Completed in pull request #102

PR #91, opened 2020-07-01
Uncertainties & File Extension Added

tamimfatahi:
Added uncertainties to be scraped from NEA for limb darkening coefficient calculations issue #16 . In order to scrape these uncertainties, you must delete the previous existing eaConf.json to scrape again. Also, .fts file extensions now added and tested on certain fits functions, but will need further testing for issue #34 . Minor efficiency change in searching for file extensions as well as formatting.

### DIFF
--- a/exotic.py
+++ b/exotic.py
@@ -140,53 +140,6 @@ def binner(arr, n, err=''):
         err = np.array([np.sqrt(1. / np.nansum(1. / (np.array(i) ** 2.))) for i in why])
         return arr, err
 
-
-# finds the planet line in the composite dictionary
-# returns -1 if its not there
-def findPlanetLineComp(planName, dataDictionary):
-    coun = 0
-    for line in dataDictionary:
-        if line['fpl_name'] == planName:
-            index = coun
-        coun = coun + 1
-    return index
-
-
-# finds the line number of the planetName in the confirmed dictionary
-# returns -1 if its not there
-def findPlanetLineConf(planName, dataDictionary):
-    coun = 0
-    index = -1
-    # account for mistakes in capitalization, spaces, and dashes
-    noSpaceP = planName.replace(" ", "")
-    noSpaceDashP = noSpaceP.replace("-", "")
-    noCapsSpaceP = noSpaceDashP.lower()
-    for line in dataDictionary:
-        exoPname = line['pl_name']
-        exoPnoSpace = exoPname.replace(" ", "")
-        exoPnoSpaceDash = exoPnoSpace.replace("-", "")
-        exoPnoSpaceLower = exoPnoSpaceDash.lower()
-        if noCapsSpaceP == exoPnoSpaceLower:
-            index = coun
-        coun = coun + 1
-    if index != -1:
-        return index
-    else:
-        # print('Error! Could Not Find Planet with the name: '+planName)
-        # sys.exit()
-        return -1
-
-
-def findPlanetLinesExt(planName, dataDictionary):
-    coun = 0
-    indexList = []  # this has multiple lines per planet so list needed
-    for line in dataDictionary:
-        if line['mpl_name'] == planName:
-            indexList.append(coun)
-        coun = coun + 1
-    return indexList
-
-
 ## ARCHIVE PRIOR SCRAPER ################################################################
 pi = 3.14159
 au = 1.496e11  # m
@@ -235,7 +188,8 @@ def new_scrape(filename="eaConf.json"):
         "select"   : "pl_name,hostname,tran_flag,pl_massj,pl_radj,pl_ratdor,"
                      "pl_orbper,pl_orbpererr1,pl_orbpererr2,pl_orbeccen,"
                      "pl_orbincl,pl_orblper,pl_tranmid,pl_tranmiderr1,pl_tranmiderr2,"
-                     "st_teff,st_tefferr1,st_met,st_meterr1,st_logg,st_loggerr1,st_mass,st_rad,ra,dec",
+                     "st_teff,st_tefferr1,st_tefferr2,st_met,st_meterr1,st_meterr2,"
+                     "st_logg,st_loggerr1,st_loggerr2,st_mass,st_rad,ra,dec",
         "from"     : "ps",  # Table name
         "where"    : "tran_flag = 1 and default_flag = 1",
         "order by" : "pl_name",
@@ -283,28 +237,38 @@ def new_scrape(filename="eaConf.json"):
 
 
 def new_getParams(data):
+
+    # Find the larger uncertainty parameter between the + (1) and - (2)
+    parameter_unc_dict = {'st_tefferr': None, 'st_meterr': None, 'st_loggerr': None}
+
+    for key in parameter_unc_dict:
+        if abs(data[key + '1']) >= abs(data[key + '2']):
+            parameter_unc_dict[key] = abs(data[key + '1'])
+        else:
+            parameter_unc_dict[key] = abs(data[key + '2'])
+
     # translate data from Archive keys to Ethan Keys
 
     planetDictionary = {
         'pName': data['pl_name'],
         'sName': data['hostname'],
-        'rprs': data['pl_radj']*rjup/(data['st_rad']*rsun),
-        'aRs': data['pl_ratdor'],
-        'midT': data['pl_tranmid'],
-        'midTUnc': data['pl_tranmiderr1'],
+        'ra': data['ra'],
+        'dec': data['dec'],
         'pPer': data['pl_orbper'],
         'pPerUnc': data['pl_orbpererr1'],
-        'flag': data['tran_flag'],
+        'midT': data['pl_tranmid'],
+        'midTUnc': data['pl_tranmiderr1'],
+        'rprs': data['pl_radj']*rjup/(data['st_rad']*rsun),
+        'aRs': data['pl_ratdor'],
         'inc': data['pl_orbincl'],
         'ecc': data.get('pl_orbeccen', 0),
         'teff': data['st_teff'],
-        'teffUnc': data['st_tefferr1'],
+        'teffUnc': parameter_unc_dict['st_tefferr'],
         'met': data['st_met'],
-        'metUnc': data['st_meterr1'],
+        'metUnc': parameter_unc_dict['st_meterr'],
         'logg': data['st_logg'],
-        'loggUnc': data['st_loggerr1'],
-        'ra': data['ra'],
-        'dec': data['dec']
+        'loggUnc': parameter_unc_dict['st_loggerr'],
+        'flag': data['tran_flag']
     }
 
     return planetDictionary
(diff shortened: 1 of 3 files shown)


exotic.py
```diff
@@ -407,30 +410,32 @@ def user_input(prompt, type_, val1=None, val2=None):
...
+                        inputfiles.append(os.path.join(directory, file))
+                # If we find files, then stop the for loop and while loop
+                if inputfiles:
+                    return directory, inputfiles
+
+            # If we don't find any files, then we need the user to check their directory and loop over again....
+            raise FileNotFoundError
+        except FileNotFoundError:
+            print("Error: " + fileName + " .FITS or .FTS files not found in " + directory + ". Please try again.")
+            directory = str(input("Enter the Directory Path where FITS or .FTS Image Files are located: "))
```
rzellem, line 438: In the future, we should ask the user to input their file extension if EXOTIC cannot automatically find it. They might use a non-standard image extension.

exotic.py
```diff
@@ -235,7 +235,7 @@ def new_scrape(filename="eaConf.json"):
         "select"   : "pl_name,hostname,tran_flag,pl_massj,pl_radj,pl_ratdor,"
                      "pl_orbper,pl_orbpererr1,pl_orbpererr2,pl_orbeccen,"
                      "pl_orbincl,pl_orblper,pl_tranmid,pl_tranmiderr1,pl_tranmiderr2,"
-                     "st_teff,st_met,st_logg,st_mass,st_rad,ra,dec",
+                     "st_teff,st_tefferr1,st_met,st_meterr1,st_logg,st_loggerr1,st_mass,st_rad,ra,dec",
```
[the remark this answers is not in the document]
tamimfatahi: After consulting with Kyle, it seems that scraping both parameters is not needed for limb darkening. If we scraped both the + and - uncertainties, it would cause a longer load time for users to have the eaConf.json on their computer (and, depending on the paper, they're usually the same). It is possible to do of course and can be implemented if needed. I'm not sure what error `3` is, but `2` is the - and `1` is the + for uncertainties. If needed for the LDTK, I could implement both `1` and `2` to find the larger of the two for limb darkening calculations.

exotic.py
tamimfatahi: I will definitely modify this portion of the code later on for accepting/searching those extensions when solving the ticket - I added this as a bandaid for those that do not want convert between .fits and .fts to run in EXOTIC. I can revert this back though and solve it at a later time if needed to complete issue #34

jpl-jengelke: I would change the matching algorithm slightly. 
```
   fits_extensions = ['.fits', '.fit', '.fts']
   ...
   if file.lower().endswith(exti.lower()): 
       ...
```
Now there is a smaller list of extensions to manage, and as well the users' files may be named with `.Fits` or `.Fit` or any case permutation.

jpl-jengelke: This probably follows up to other code segments, too. After the FileNotFoundError (and any other blocks) it would be helpful to have a fall through exception, like `OSError` to catch any local I/O Error that may occur. [According to the latest Python docs, `OSError` is basically at the root of file system failures.](https://docs.python.org/3/library/exceptions.html#OSError)

exotic.py
```diff
@@ -140,53 +140,6 @@ def binner(arr, n, err=''):
...
-def findPlanetLinesExt(planName, dataDictionary):
-    coun = 0
-    indexList = []  # this has multiple lines per planet so list needed
-    for line in dataDictionary:
-        if line['mpl_name'] == planName:
-            indexList.append(coun)
-        coun = coun + 1
-    return indexList
-
-
```
tamimfatahi, line 189: This seems to be the old way of searching for an exoplanet name in the NEA - removed it as we updated the API.

exotic.py
```diff
@@ -288,20 +242,26 @@ def new_getParams(data):
...
+        'teffUncPos': data['st_tefferr1'],
+        'teffUncNeg': data['st_tefferr2'],
+        'met': data['st_met'],
+        'metUncPos': data['st_meterr1'],
+        'metUncNeg': data['st_meterr2'],
+        'logg': data['st_logg'],
+        'loggUncPos': data['st_loggerr1'],
+        'loggUncNeg': data['st_loggerr2'],
+        'flag': data['tran_flag']
     }
```
tamimfatahi, line 265: Scraping both the + and - uncertainties for temperature, surface gravity, and metallicity.

exotic.py
```diff
@@ -288,20 +242,26 @@ def new_getParams(data):
...
+        'teff': data['st_teff'],
+        'teffUncPos': data['st_tefferr1'],
+        'teffUncNeg': data['st_tefferr2'],
+        'met': data['st_met'],
+        'metUncPos': data['st_meterr1'],
+        'metUncNeg': data['st_meterr2'],
+        'logg': data['st_logg'],
+        'loggUncPos': data['st_loggerr1'],
+        'loggUncNeg': data['st_loggerr2'],
+        'flag': data['tran_flag']
```
tamimfatahi, line 264: Scraping both the + and - uncertainties for temperature, surface gravity, and metallicity.

exotic.py
```diff
@@ -407,30 +367,108 @@ def user_input(prompt, type_, val1=None, val2=None):
...
+                    pDict['ra'] = astropy.coordinates.Angle(raStr + " hours").deg
+            elif key == 'dec':
+                    decStr = input('Enter the ' + planet_params[i] + ': ')
+                    decStr = decStr.replace(' ', '').replace(':', '')
+                    pDict['dec'] = astropy.coordinates.Angle(decStr + " degrees").deg
+            else:
+                pDict[key] = user_input('Enter the ' + planet_params[i] + ': ', type_=float)
+
+    return pDict
+
```
tamimfatahi, line 434: Added this portion of the code to make life easier when checking against the NEA or user input. Saved coding space and will make it easier to write to the parameters to init.txt file later on.

exotic.py
```diff
@@ -407,30 +367,108 @@ def user_input(prompt, type_, val1=None, val2=None):
...
+            # If the file or directory does not exist
+            else:
+                raise OSError
+
+        except FileNotFoundError:
+            print("Error: " + fileName + " files not found with .fits, .fit or .fts extensions in " + directory + ". Please try again.")
+            directory = input("Enter the directory path where " + fileName + " files are located: ")
+        except OSError:
+            print("Error: No such directory exists. Please try again.")
+            directory = input("Enter the directory path where " + fileName + " files are located: ")
```
tamimfatahi, line 471: Extensions checking for imaging files updated w/ new suggestions.

exotic.py
```diff
@@ -889,7 +926,7 @@ def contextupdt(times=None, airm=None):
 
 # -- LIGHT CURVE MODEL -- ----------------------------------------------------------------
 def lcmodel(midTran, radi, am1, am2, theTimes, theAirmasses, plots=False):
-    sep, ophase = time2z(theTimes, inc, midTran, semi, planetPeriod, eccent)
+    sep, ophase = time2z(theTimes, pDict['inc'], midTran, pDict['aRs'], pDict['pPer'], pDict['ecc'])
```
tamimfatahi, line 929: pDict will now be scattered throughout the code rather than creating a new variable for each one. These remain static and already exist, no need to create new variables.

tamimfatahi:
My apologies found an error. Let me re-review and I'll continue commenting the changes.

exotic.py
```diff
@@ -1096,16 +1134,20 @@ def parse_args():
...
-        while carryOn != 'continue':
-            carryOn = input('Type continue after the first image has been taken and saved: ')
+        while True:
+            try:
+                carryOn = input('Type continue after the first image has been taken and saved: ')
+                if carryOn != 'continue':
+                    raise ValueError
+                break
+            except ValueError:
+                continue
```
tamimfatahi, line 1150: Pythonic correction.

exotic.py
```diff
@@ -1330,31 +1385,43 @@ def parse_args():
...
+                        raise ValueError("You forgot the sign for the longitude! East is '+' and West is '-'. Please try again.")
+                    longit = float(longitStr)
+                    if longit <= -180.00 or longit >= 180.00:
+                        raise ValueError('Your longitude is out of range. Please enter a longitude between -180 and +180 (deg)')
+                    break
+                # check to make sure they have a sign
+                except ValueError as err:
+                    print(err.args)
+                    longitStr = input("Enter the longitude of where you observed (deg) "
+                                      "(Don't forget the sign where East is '+' and West is '-'): ")
```
tamimfatahi, line 1424: Pythonic correction w/ extra validation check.

exotic.py
```diff
@@ -1314,11 +1368,12 @@ def parse_args():
...
-                        CandidatePlanetBool = True
+                        CandidatePlanetBool = False
                         break
-            idx = planets.index(targetName.lower().replace(' ', '').replace('-', ''))
-            pDict = new_getParams(data[idx])
-            print('\nSuccessfuly found ' + targetName + ' in the NASA Exoplanet Archive!')
+            if CandidatePlanetBool:
+                idx = planets.index(targetName.lower().replace(' ', '').replace('-', ''))
+                pDict = new_getParams(data[idx])
+                print('\nSuccessfuly found ' + targetName + ' in the NASA Exoplanet Archive!')
```
tamimfatahi, line 1376: issue #96

exotic.py
```diff
@@ -1301,10 +1354,11 @@ def parse_args():
         # t.start()
         # check to make sure the target can be found in the exoplanet archive right after they enter its name
 
-        if not os.path.exists("eaConf.json"):
+        # Checks to see if the file exists or is over one week old to scrape/rescrape parameters (units in seconds)
+        if not os.path.exists("eaConf.json") or time.time() - os.path.getmtime('eaConf.json') > 604800:
```
tamimfatahi, line 1358: Scraping will now be time-based. If planetary parameters have been on desktop for a week, rescrape.

exotic.py
```diff
@@ -1281,9 +1324,19 @@ def parse_args():
...
+            try:
+                # In case the user forgets the trailing / for the folder
+                if saveDirectory[-1] != "/":
+                    saveDirectory += "/"
+                if os.path.isdir(saveDirectory):
+                    break
+                raise OSError
+            except OSError:
+                print('Error: the directory entered does not exist. Please try again.')
+                saveDirectory = input("Enter the Directory to Save Plots into: ")
```
tamimfatahi, line 1339: issue #97

exotic.py
```diff
@@ -1143,19 +1185,20 @@ def parse_args():
...
                     with open(initfilename, 'r') as f:
                         inits = f.readlines()
-                        break
+                    break
                 except FileNotFoundError:
-                    print("Initialization file not found. Please try again.")
-                    initfilename = str(input("\nPlease enter the Directory and Filename of your Initialization File: "))
+                    print("Error: Initialization file not found. Please try again.")
+                except IsADirectoryError:
+                    print('Error: Entered a directory. Please try again.')
```
tamimfatahi, line 1201: Added an additional validation check to inform user they entered a directory instead of a file.

exotic.py
```diff
@@ -1470,167 +1537,11 @@ def parse_args():
...
-            print('Limb Darkening Coefficients')
-            print('***************************')
-            # stellar temperature
-            starTeff = user_input("\nEnter the host star's effective temperature (K): ", type_=float)
-
-            # metallicity
-            starMetall = user_input("\nEnter the host star's metallicity ([Fe/H]): ", type_=float)
-
-            # Log g
-            starSurfG = user_input("\nEnter the host star's surface gravity (log(g)): ", type_=float)
```
tamimfatahi, line 1633: Deleted as replaced by new parameter check function from above.

requirements.txt
```diff
@@ -9,3 +9,6 @@ pymc3==3.8
 python_dateutil==2.8.1
 scipy==1.4.1
 theano==1.0.4
+skimage==0.15.0
+semantic-version==2.8.5
+LDTk==1.1.1
```
tamimfatahi, line 14: Packages needed for LDTk. skimage has to be 0.15.0 for now. Needs to be updated later to 0.17.2 as EXOTIC gets warning with the updated package.

exotic.py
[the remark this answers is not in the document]
rzellem: I would implement both `1` and `2` and then take the geometric mean, a la EXCALIBUR.

exotic.py
```diff
@@ -407,30 +367,108 @@ def user_input(prompt, type_, val1=None, val2=None):
             return option
 
 
-def check_file_extensions(directory, fileName):
-    # Add / to end of directory if user does not input it
-    if directory[-1] != "/":
-        directory += "/"
```
rzellem, line 413: why is this now removed?

exotic.py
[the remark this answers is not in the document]
tamimfatahi: Taking the geometric mean is being done in my separate environment and will be added to EXOTIC's LD calculations function.

exotic.py
tamimfatahi: Not removed, it's lower down. Due to the increase in code, it moved it down on the right-hand side.

rzellem: Ah nevermind- you've tweaked it below.

exotic.py
rzellem: Great idea!

exotic.py
tamimfatahi: I mixed up the True and False for candidate planets. Will revert back once updated with new planetary suggestions.

rzellem: I think you might have mixed up what a `CandidatePlanet` is - this is a non-confirmed planet, so it is still a "candidate". So if `targetName == 'manual'` I think `CandidatePlanetBool = True` instead. Correct?

exotic.py
```diff
@@ -1330,31 +1384,43 @@ def parse_args():
...
-                latiStr = str(input("Enter the latitude of where you observed (deg) (Don't forget the sign where North is '+' and South is '-'): "))
-                noSpaceLati = latiStr.replace(" ", "")
-                latiSign = noSpaceLati[0]
-            lati = float(latiStr)
-
-            # handle longitude
+                latiStr = input("Enter the latitude of where you observed (deg) "
+                                "(Don't forget the sign where North is '+' and South is '-'): ")
+            # Latitude
+            while True:
```
rzellem, line 1392: just a question- doesn't this while loop always run since it is always true? If so, why use this coding? (just curious!)

tamimfatahi: There's a break at the end. If it passes all my parameters, it will break. If it doesn't, it will raise an error and loop over again.

exotic.py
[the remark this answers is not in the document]
rzellem: Ok great thanks.

exotic.py
tamimfatahi: Updated with your suggestion John to the following:

`
              if os.path.isdir(directory):
                 # Loop until we find something
                 for exti in file_extensions:
                     for file in os.listdir(directory):
                         if file.lower().endswith(exti.lower()):
                             inputfiles.append(os.path.join(directory, file))
                     # If we find files, then stop the for loop and while loop
                     if inputfiles:
                         return directory, inputfiles

                  # If we don't find any files, then we need the user to check their directory and loop over again
                 if not inputfiles:
                     raise FileNotFoundError

              # If the directory does not exist
             else:
                 raise OSError

          except FileNotFoundError:
             print("Error: " + fileName + " files not found with .fits, .fit or .fts extensions in " + directory + ". Please try again.")
             directory = input("Enter the directory path where " + fileName + " files are located: ")
         except OSError:
             print("Error: No such directory exists. Please try again.")
             directory = input("Enter the directory path where " + fileName + " files are located: ")
`

jpl-jengelke:
Also, one could ask if they would like to create the directory.

exotic.py
jpl-jengelke: It may be good to ask if they would like to create the directory if it doesn't exist.

exotic.py
tamimfatahi: Went ahead and fixed it and waiting on your opinion of planet checking.

exotic.py
```diff
@@ -407,30 +367,123 @@ def user_input(prompt, type_, val1=None, val2=None):
...
+    try:
+        directoryName = input('Enter the name for your new directory: ')
+        newDirectoryPath = os.getcwd()
+        saveDirectory = newDirectoryPath + '/' + directoryName + '/'
+        os.mkdir(saveDirectory)
+    except OSError:
+        print("Creation of the directory %s failed" % saveDirectory)
+    else:
+        print("Successfully created the directory %s " % saveDirectory)
+        return saveDirectory
```
tamimfatahi, line 381: New directory creation function added for saving plots and anything in the future needed.

Issue #115, opened 2020-07-09
Update scikit-image package 


tamimfatahi:
Currently, EXOTIC works well on scikit-image==0.15.0, but throws an error on newer editions. Update EXOTIC to work for the latest package.

tamimfatahi:
![Register Translation](https://user-images.githubusercontent.com/32148306/87101558-4ca8a580-c204-11ea-8136-9bf52a29fef8.PNG)

tamimfatahi:
Completed in pull request #116

PR #116, opened 2020-07-10
Issue_115: scikit-image==0.17.2 update for EXOTIC and Candidate Planet Bug (no ticket)

tamimfatahi:
issue #115 This update is going to require all users to update this package since this function does not exist in 0.15.0. Please let them know since I'm sure it will come up as an error for some! Candidate planet bug fixed. Should work smoothly now.


tamimfatahi:
Finding more errors. Stand-by for last push.

PR #117, opened 2020-07-10
Issue 93: directoryP referenced without being initialized

mjs2369:
Fixing initialization error & incorrect warning on first iteration of while loop

### DIFF
--- a/exotic.py
+++ b/exotic.py
@@ -981,6 +981,12 @@ def realTimeReduce(i):
 
     # -------TIME SORT THE FILES--------------------------------------------------------------------------------
     directoryP = ""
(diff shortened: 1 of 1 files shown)


exotic.py
```diff
@@ -1164,6 +1171,7 @@ def parse_args():
         print('**************************************************************\n')
 
         directToWatch = str(input("Enter the Directory Path of imaging files: "))
+        directoryP = ""
```
rzellem, line 1174: huh- wouldn't think this is necessary, but it is a nice safety net just in case

exotic.py
```diff
@@ -1205,6 +1213,8 @@ def parse_args():
         print('Complete Reduction Routine')
         print('**************************\n')
 
+        directoryP = ""
```
rzellem, line 1216: same thing here too

PR #124, opened 2020-07-13
Plugged FITS loading memory leak.

rzellem:
Thanks to @jpl-jengelke for figuring out how to fix this problem!

### DIFF
--- a/exotic.py
+++ b/exotic.py
@@ -958,7 +958,7 @@ def realTimeReduce(i):
     # sorts the times for later plotting use
     sortedTimeList = sorted(timeList)
 
-    hdul = fits.open(timeSortedNames[0])  # opens the fits file
+    # hdul = fits.open(name=timeSortedNames[0], memmap=False, cache=False, lazy_load_hdus=False)  # opens the fits file
     # Extracts data from the image file and puts it in a 2D numpy array: firstImageData
     firstImageData = fits.getdata(timeSortedNames[0], ext=0)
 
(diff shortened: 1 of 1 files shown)


exotic.py
```diff
@@ -972,11 +972,15 @@ def realTimeReduce(i):
...
     for imageFile in timeSortedNames:
 
-        hDul = fits.open(imageFile)  # opens the fits file
+        hDul = fits.open(name=imageFile, memmap=False, cache=False, lazy_load_hdus=False)  # opens the fits file
         # Extracts data from the image file and puts it in a 2D numpy array: imageData
         imageData = fits.getdata(imageFile, ext=0)
         header = fits.getheader(imageFile)
 
+        hdul.close()
+        del hdul[0].data
```
pearsonkyle, line 981: Why no `del(hdul)` to delete the data and the header?

rzellem: Good point. Changing to `del hdul` instead.

jpl-jengelke: Note the case of the variables. ... `hDul` versus `hdul` in the `del` statement. There won't be a need for `hdul = None` if `del hdul` is used. 

On another note, the same issue is apparent in several other areas in the code. where `fits.open(...)` is utilized.

jpl-jengelke:
Not sure this was published, so reposting it here. ... 

Note the case of the variables. ... `hDul` versus `hdul` in the `del` statement. There won't be a need for `hdul = None` if `del hdul` is used.

On another note, the same issue is apparent in several other areas in the code. where `fits.open(...)` is utilized.

rzellem:
Addressed comments from both @pearsonkyle and @jpl-jengelke in latest push. Please review and approve.

rzellem:
Addressed @jpl-jengelke 's concerns. Using superpowers to merge.

rzellem:
Opening back up again because @rzellem found a bug....

Issue #13, opened 2020-06-12
Check User Inputs


rzellem:
Move planetary system parameters into the init.txt file. Have EXOTIC then check these values against the NASA Exoplanet Archive and raise an exception if they do not match. If the user does not supply them, then proceed as usual by scraping them from the Archive automatically and then asking user for agreement.

Also double check that the x,y pixel coordinates of the target star agree with the plate solution of the image.

tamimfatahi:
I can check the x,y pixel coordinates of the target star on this issue. I can tag team this

tamimfatahi:
Completed in pull request #128

PR #128, opened 2020-07-14
Issue 13: Checking target star's pixel coordinates 

tamimfatahi:
Checking the user entered pixel coordinates for the target star for issue #13 . Also entered a scraper, but mainly to save it on Github. Works, but needs the correct inits.txt and formatting is off. Not called in main function and will not affect the code.


exotic.py
```diff
@@ -1590,163 +1711,177 @@ def parse_args():
...
+
+                # Image Alignment
+                print("\nAligning your images from .FITS. Please wait.")
+                done = False
+                t = threading.Thread(target=animate, daemon=True)
+                t.start()
+                sortedallImageData, unalignedBoolList, boollist = image_alignment(sortedallImageData)
+                done = True
+                print('\n\nImages Aligned.')
+                break
```
tamimfatahi, line 1884: The idea behind tracking the target pixel coordinates is to do it by these steps: time sort, plate solution, pixel coordinates, then image alignment. I did image alignment last because I wanted to check the pixel coordinates to save time for the user having to do image alignment twice.

exotic.py
```diff
@@ -1590,163 +1711,177 @@ def parse_args():
...
+                    hdulWCS = fits.open(name=wcsFile, memmap=False, cache=False, lazy_load_hdus=False)  # opens the fits file
+
+                    # Save previously entered x and y pixel coordinates before checking against plate solution
+                    saveUIprevTPX, saveUIprevTPY = UIprevTPX, UIprevTPY
+                    UIprevTPX, UIprevTPY = check_targetpixelwcs(UIprevTPX, UIprevTPY, pDict['ra'],
+                                                                pDict['dec'], hdulWCS)
+
+                    # If the coordinates were not changed, do not loop over again
+                    if UIprevTPX != saveUIprevTPX or UIprevTPY != saveUIprevTPY:
+                        continue
```
tamimfatahi, line 1874: If either target coordinate changes, it will loop over again (at most 2 times).

exotic.py
```diff
@@ -605,13 +684,50 @@ def plate_solution(fits_file, saveDirectory):
...
+                print('\nError: The X Pixel Coordinate entered does not match the right ascension.')
+                raise ValueError
+            if expdec - 0.00556 >= obsdec or obsdec >= expdec + 0.00556:
+                print('\nError: The Y Pixel Coordinate entered does not match the declination.')
+                raise ValueError
+            return pixx, pixy
+        except ValueError:
+            pixx = user_input("Please re-enter the target star's X Pixel Coordinate: ", type_=int)
+            pixy = user_input("Please re-enter the target star's Y Pixel Coordinate: ", type_=int)
+
```
tamimfatahi, line 717: Pixel coordinates will have to be fixed here, not allowing the user to re-enter the wrong coordinate. This makes the while loop only iterate twice.

rzellem: Few comments:

1. Why does the code ask the user to input the target star's x/y pixel location when it already *knows* the location of the target/comp star? Since you've done a WCS solution to the image, every pixel is registered to a RA/DEC location, so EXOTIC can exactly find by itself the location of the target/comps on the focal plane without any user input.

2. What happens if the star has moved a lot (due to proper motion) and is not actually at the pixel location that EXOTIC thinks? In other words, you should raise an error and give the user the option to proceed with their pixel coordinates or to re-enter. Right now, you only let the user re-enter new pixel values.

exotic.py
```diff
@@ -1590,163 +1711,177 @@ def parse_args():
...
-                        #     # sortedallImageData[idx,xi,yi] = np.nanmedian(sortedallImageData[:,xi,yi])
-                        #     idx = np.abs(sortedallImageData[:,xi,yi]-np.nanmedian(sortedallImageData[:,xi,yi])) >  5*np.nanstd(sortedallImageData[:,xi,yi])
-                done = True
-
-            # if len(sortedTimeList) == 0:
-            #     print("Error: .FITS files not found in " + directoryP)
-            #     sys.exit()
-
-            # -------OPTIMAL COMP STAR, APERTURE, AND ANNULUS CALCULATION----------------------------------------
+            # Loop placed to check user-entered x and y target coordinates against WCS. Should iterate at MAX 2 times.
```
rzellem, line 1714: Why only integrating a max of 2 times?

exotic.py
rzellem: Why 2 times?

exotic.py
tamimfatahi: 1. Although EXOTIC does know where the ra/dec is for the target stars due to having all the pixels converted from another function I decommissioned, I thought when we discussed not having to automate everything, this was one of those things. I can automate it if need be to find the locations if the user entered them incorrectly. 

2. Yeah, for now it is hardstuck on having them re-enter if it doesn't deem their pixel coordinates to be in the same ra/dec. The reason why is because it goes off the 'firstimage' that it sees it can fit a centroid to. It gives the user a downloaded file of the newfits.fits to look over it and re-enter the coordinates. I can make this optional of course since I can see it running into issues.

exotic.py
tamimfatahi: If the user entered the target coordinates correct the first time, it will loop once. If the user entered the target coordinates incorrectly, it will loop over again once it hits the check_targetpixelcoords() function and have the user enter in the correct ones. Once that happens, the time sorting and plate solution will be redone.

exotic.py
rzellem: 1. I would have EXOTIC perhaps suggest the pixel x/y coordinates that it *thinks* the target star is. That way the user can check their values against EXOTIC and have the option to either: 1. keep their original x/y position, 2. adopt EXOTIC's suggested values, or 3. enter completely new x/y pixel coords.

2. I would still add in the option for the user to use their originally-entered pixel coordinates, but point the user to the WCS-solved and aligned FITS file to double check.

tamimfatahi: That sounds great to me!

tamimfatahi:
Changes done - up for review when ready.

exotic.py
tamimfatahi: Corrected to do your 1. and 2.

inits.json
```diff
@@ -0,0 +1,54 @@
+{
+    "user_info": [
+        {
+            "Directory with FITS files": "./sample-data/HatP32Dec202017",
+            "Directory to Save Plots": "./sample-data",
+            "Directory of Flats": "n",
```
rzellem, line 6: I'm a little worried about having the answer be "n" here. Maybe we should make the code more robust to accept a "n/a" or "none" or give the user instructions for what to do if they do not have cals? Can we add a comment in-line after the `,`? I.e., can we put in `# please put 'n' if no calibrations` - I'm not sure if this will work with the json file format or not

inits.json
```diff
@@ -0,0 +1,54 @@
...
+            "Directory of Darks": "n",
+            "Directory of Biases": "n",
+
+            "AAVSO Output?": "y",
```
rzellem, line 10: I'm thinking that we should _always_ do the AAVSO output and then if the user does not have an AAVSO Observing Account Number, they just put `none` or something else applicable. That way they can have the output file already on their computer and if they reconsider and want to upload to AAVSO, they don't have to re-run EXOTIC again.

inits.json
rzellem: ^ definitely address this comment

inits.json
rzellem: ^ this can be a new opened ticket for a future release

inits.json
```diff
@@ -0,0 +1,54 @@
...
+            "AAVSO Observer Account Number": "RTZ",
+            "Secondary Observer Codes": "n",
+
+            "Observation date": "December 17, 2017",
+            "Obs. Latitude (+=N,-=S)": "+31.68",
```
rzellem, line 15: We might want to put in a comment the format of the Lat and Long for the user.

inits.json
```diff
@@ -0,0 +1,54 @@
...
+            "Obs. Longitude (+=E,-=W)": "-110.88",
+            "Obs. Elevation (meters)": 2616,
+
+            "Camera Type (CCD or DSLR)": "CCD",
+            "Pixel Binning": "1x1",
+            "Exposure Time (seconds)": "60",
```
rzellem, line 21: So why is this value in quotations while the `Obs. Elevation` is not? They are both floats, so they shouldn't need the quotation marks.

Also...we should consider just grabbing the exposure time from the FITS header. That way we can do a quick check to make sure that the user has all of their images at the same exposure time. If not, then we can scale the images to the same exposure time, assuming n_photons ~= n_time

inits.json
```diff
@@ -0,0 +1,54 @@
...
+            "Filter Name (aavso.org/filters)": "V",
+            "Observing Notes": "Weather, seeing was nice.",
+
+            "Target Star X & Y Pixel": [424, 286],
+            "Comparison Star(s) X & Y Pixel": [[465, 183], [512, 263]]
+        }
+    ],
+    "planetary_parameters": [
+        {
+            "Planet's Name": "HAT-P-32 b",
```
rzellem, line 31: should we give the user some guidance here that their star name must match how it is listed in the NASA Exoplanet Archive?

inits.json
rzellem: Looks like JSONs do not support comments, but we can get around that by doing something like this: https://stackoverflow.com/questions/244777/can-comments-be-used-in-json

exotic.py
```diff
@@ -371,111 +376,49 @@ def user_input(prompt, type_, val1=None, val2=None):
...
-        newDirectoryPath = os.getcwd()
-        saveDirectory = newDirectoryPath + '/' + directoryName + '/'
-        os.mkdir(saveDirectory)
+        directoryname = input('Enter the name for your new directory: ')
+        newdirectoryPath = os.getcwd()
+        savedirectory = newdirectoryPath + '/' + directoryname + '/'
+        os.mkdir(savedirectory)
     except OSError:
-        print("Creation of the directory %s failed" % saveDirectory)
+        print("Creation of the directory %s failed" % savedirectory)
```
rzellem, line 384: If the code is unable to create the directory, then this function will fail and the directory will not be created, potentially crashing EXOTIC. Maybe you should use a `while` loop here like you've done with other checks?

exotic.py
```diff
@@ -1637,26 +1551,31 @@ def parse_args():
...
             else:
                 AAVSOBool = True
                 # userNameEmails = str(input('Please enter your name(s) and email address(es) in the format: Your Name (youremail@example.com), Next Name (nextemail@example.com), etc.  '))
-                userCode = str(input('Please enter your AAVSO Observer Account Number: '))
-                secuserCode = str(input('Please enter your comma-separated secondary observer codes (or type none if only 1 observer code): '))
-            cameraType = str(input("Please enter your camera type (CCD or DSLR): "))
-            binning = str(input('Please enter your pixel binning: '))
-            exposureTime = str(input('Please enter your exposure time (seconds): '))
-            filterName = str(input('Please enter your filter name from the options at '
+                infoDict['aavsonum'] = str(input('Please enter your AAVSO Observer Account Number: '))
```
rzellem, line 1560: Recommend adding `(please input n/a if you do not currently have an account)` as we should always build the AAVSO file from now on

inits.json
tamimfatahi: If the exoplanet is in the NEA, it already scrapes the planet and star names from the NEA and ignores the inits.json inputs. The values when using an exoplanet in NEA that EXOTIC automatically uses are the planet name, star name, right ascension, and declination. All other parameters will be taken from inits.json. If this is a candidate planet, it will use all the inits.json parameters.

exotic.py
rzellem: Right, but if a user defines their xy pixels on the first image but the 10th image is chosen as the one that all others are aligned, there is a chance (albeit likely small...hopefully) that the image will shift enough on the focal plane that the user will select the "wrong" star. But perhaps your checking code will help prevent this.

For now, I would tell the user what image we are using as the "first" one (both the image number and the full filename)- that way they can open that FITS file later to confirm the star's xy pixel coordinates.

inits.json
tamimfatahi: Exposure times will be opened in a separate ticket as there needs to be more tests with different headers and how they spell label exposure.

exotic.py
tamimfatahi: Done in the newest push. Created a file with the original name and gives the image number.

inits.json
tamimfatahi: Done in this push. Users must have to put in their AAVSO code or N/A if none.

inits.json
tamimfatahi: Left in a comment for how to format latitude and longitude.

inits.json
tamimfatahi: RA and DEC values are now matched to the NEA for comparison and the user can choose between the two or enter a new one.

exotic.py
tamimfatahi: Done!

exotic.py
tamimfatahi: Done!

tamimfatahi:
Most suggestions are completed. If not, a ticket was open. Highly recommend everyone tests this as much as possible as there was a lot changed regarding the arrangement of dictionary values that are used all over the code.

tamimfatahi:
New pull request will come in to fix candidate planet bug.

PR #145, opened 2020-07-16
Issue 144: hdul bug with no plate solution

tamimfatahi:
hdul bug for searching PIXSCALE in FITS header. Also added a Python version check to EXOTIC.

### DIFF
--- a/exotic.py
+++ b/exotic.py
@@ -40,6 +40,8 @@ import threading
 import time
 import sys
 
+print('Python Version: %s' % sys.version)
+
 # To increase memory allocation for EXOTIC; allows for more fits files
 # import resource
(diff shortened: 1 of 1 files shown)


exotic.py
```diff
@@ -2145,7 +2147,7 @@ def lc2min(x):
                 imscalen = imageheader['IM_SCALE']
                 imscaleunits = imageheader.comments['IM_SCALE']
                 imscale = imscaleunits + ": " + str(imscalen)
-            elif "PIXSCALE" in hdul[0].header:
+            elif "PIXSCALE" in imageheader:
                 imscalen = imageheader['PIXSCALE']
```
tamimfatahi, line 2151: 4 lines above we use imageheader, might as well use it again!

PR #159, opened 2020-07-21
Resolves #157 -- Separate timeout specs added for connection and response

### LINKED ISSUE #157: TAP Query timeout unreasonably low for diverse public network topologies
On line `174` in EXOTIC, they timeout is set to 90 seconds. This does not account for slower connections over public networks.

```
response = requests.get(uri_full, timeout=90)
```

This does not negate the need for a more efficient query as outlined in #156.

jpl-jengelke:
Timeout tuple added to address public connection and response timeouts on diverse networks. Also, added property section at top of file.

### DIFF
--- a/exotic.py
+++ b/exotic.py
@@ -124,6 +124,22 @@ from occultquad import *
 # time.sleep(10)
 done = True
 
+# ################### START PROPERTIES ########################################
+# CONFIGURATIONS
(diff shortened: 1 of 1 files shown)


jpl-jengelke:
Please note the closing terminology in the subject. This will automatically close the ticket when the merge is approved.

PR #191, opened 2020-07-29
Binner Bug Fix

tamimfatahi:


tamimfatahi:
Also added in bug fix for FITS file path check.

PR #204, opened 2020-07-30
Issue 197 & 195: FITS file check bug & Aperture bug

tamimfatahi:
Also left pdb to check user's issues with plate solution having trouble decoding the plate solution request from astrometry.net. Changed how the image alignment print as well.


exotic.py
```diff
@@ -572,7 +572,7 @@ def check_file_extensions(directory, filename):
                 # Loop until we find something
                 for exti in file_extensions:
                     for file in os.listdir(directory):
-                        if file.lower().endswith(exti.lower()):
+                        if file.lower().endswith(exti.lower()) and file[0:2] not in ('ref', 'new'):
```
rzellem, line 575: I'd be a little careful here- the reference file you make for the WCS coords might not be in index [0:2] depending on how the user begins the names of their FITS files

tamimfatahi: I hard-coded the reference file to always have the first 3 letters to be 'ref' and the plate solution to be 'new'

exotic.py
```diff
@@ -759,12 +759,17 @@ def plate_solution(fits_file, saveDirectory):
...
+        r = requests.post(default_url + 'upload', files=files, data=headers)
 
-    # Saves submission id for checking on the status of image uploaded
-    sub_id = r.json()['subid']
-    submissions_url = 'http://nova.astrometry.net/api/submissions/%s' % sub_id
+        # Saves submission id for checking on the status of image uploaded
+        sub_id = r.json()['subid']
+        submissions_url = 'http://nova.astrometry.net/api/submissions/%s' % sub_id
+    except:
+        print("Hello! Please head over to the Exoplanet Watch Slack Channel and ask for help before going further.")
```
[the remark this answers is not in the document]
tamimfatahi: Done!

Issue #108, opened 2020-07-09
Plate Solution Safeguard & Added Functionality 


tamimfatahi:
If nova.astrometry.net throws an error when requesting a session (such as website being down), inform the user and continue on with reducing data. Also, look into ASTAP since users do plate solve through there and see how well it works with EXOTIC when searching for coordinates. If it works, great! If not, add the ability to read off the file's information.

PR #210, opened 2020-08-03
Bug when catching exceptions for rp/rs uncertainties and aperture_min…

tamimfatahi:
… variable fixed


tamimfatahi:
Added in a possible fix for issue #108

PR #225, opened 2020-08-05
fixing syntax error

mjs2369:


exotic.py
```diff
@@ -305,7 +305,9 @@ def getJulianTime(hdul):
     imageheader = hdul[0].header
 
     exp = imageheader.get('EXPTIME')  #checking for variation in .fits header format
```
rzellem, line 307: In the future you could also code it up like this:

```
try:
    exp = imageheader.get('EXPTIME')
else:
    exp = imageheader.get('EXPOSURE')

Issue #233, opened 2020-08-07
Exception if user does not enter directory


jpl-jengelke:
If a user hits <Enter> when prompted for a directory after selecting (1) for Real Time Reduction, an exception is generated.

```
dhcp-137-79-237-151:EXOTIC johnengelke$ python3 exotic/exotic.py 
Importing Python Packages - please wait.
Python Version: 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 16:52:21) 
[Clang 6.0 (clang-600.0.57)]
Thinking /

*************************************************************
Welcome to the EXOplanet Transit Interpretation Code (EXOTIC)
Version  0.14.4
*************************************************************

Enter "1" for Real Time Reduction or "2" for for Complete Reduction: 1

**************************************************************
Real Time Reduction ("Control + C"  or close the plot to quit)
**************************************************************

Enter the Directory Path of imaging files: 
Traceback (most recent call last):
  File "exotic/exotic.py", line 1335, in <module>
    directToWatch, inputfiles = check_file_extensions(directToWatch, 'imaging')
  File "exotic/exotic.py", line 618, in check_file_extensions
    if directory[-1] != "/":
IndexError: string index out of range
dhcp-137-79-237-151:EXOTIC johnengelke$ 
```
Potential solutions: 
- Set <Enter> to current working directory
- Prompt user for actual value
- Exit with error gracefully if directory is not set.
- Set to `/` if <Enter pressed>

PR #235, opened 2020-08-07
Issue 233: FITS file directory bug fix

tamimfatahi:
Issue #233

### DIFF
--- a/exotic.py
+++ b/exotic.py
@@ -593,6 +593,7 @@ def get_planetary_parameters(candplanetbool, userpdict, pdict=None):
     return userpdict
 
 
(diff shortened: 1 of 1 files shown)


jpl-jengelke:
It looks like it solves the reported directory missing bug. I don't have any FITS files to test, but I also didn't get the error again.

tamimfatahi:
Need adjustments actually. I just tested for full reduction and noticed that the files were not being placed inside the folder for save plots.

tamimfatahi:
Okay, all set for review again. Minor tweak in saving the variable.

PR #239, opened 2020-08-11
Bug fix regarding file paths in Windows

tamimfatahi:


tamimfatahi:
If someone could test this who has Windows before pushing to main for me, that would be great!

PR #330, opened 2020-09-04
new features

pearsonkyle:
- improved sigma clipping with Savitsky Golay filter instead of rolling median
- Changed the phase calculation + corresponding errors 
- removed astropy warning from triggering corrupt file catcher
- added diagnostic plot for background estimate
- added algorithm to remove hot pixels before image alignment
- added diagnostic plot for hot pixel mask
-fixed logic if PSF fit is the best
- changed bounds for imshow in FOV plot


exotic/exotic.py
```diff
@@ -391,8 +393,8 @@ def getJulianTime(hdul):
 
 # Method that gets and returns the current phase of the target
 def getPhase(curTime, pPeriod, tMid):
-    phase = (curTime - tMid) / pPeriod
-    return phase - int(np.nanmin(phase))
+    phase = (curTime - tMid -0.5*pPeriod) / pPeriod % 1
+    return phase - 0.5
```
rzellem, line 397: why are you doing it like this?

exotic/exotic.py
```diff
@@ -2280,8 +2357,7 @@ def main():
                             'a2': 0,             #Flux lower bound
                         }
 
-                        # phase = (arrayTimes[~filtered_data]-prior['tmid'])/prior['per']
-                        phase = getPhase(arrayTimes[~filtered_data], prior['per'], prior['tmid'])
+                        phase = (arrayTimes[~filtered_data]-prior['tmid'])/prior['per']
```
rzellem, line 2360: why not call the function instead of redefining things here?

exotic/exotic.py
```diff
@@ -2303,7 +2379,7 @@ def main():
                                 elif pDict[k] == 0 or np.isnan(pDict[k]):
                                     print(" WARNING! {} uncertainty is 0. Please use a non-zero value in inits.json".format(k))
                                     pDict[k] = 1
-                            elif not pDict[k]:
+                            elif pDict[k] == None:
                                 print(" WARNING! {} is None. Please use a numeric value in inits.json".format(k))
```
rzellem, line 2383: please give the user the option to add it at the command line--- that way they don't have to go back and restart the code after editing the inits.json file

exotic/exotic.py
```diff
@@ -2340,18 +2416,28 @@ def main():
...
                             minAnnulus = annulusR  # then set min aperature and annulus to those values
                             minAperture = apertureR
-                            # gets the centroid trace plots to ensure tracking is working
-                            finXTargCentArray = np.array(xTargCent)
-                            finYTargCentArray = np.array(yTargCent)
-                            finXRefCentArray = np.array(xRefCent)
-                            finYRefCentArray = np.array(yRefCent)
+
+                            if len(xTargCent) > 0:
+                                # gets the centroid trace plots to ensure tracking is working
```
rzellem, line 2421: how does this work given that Tamim aligns all of the images up?

exotic/exotic.py
```diff
@@ -2457,7 +2543,9 @@ def main():
             if minAperture >= 0:
                 ref_circle = plt.Circle((finXRefCent[0], finYRefCent[0]), minAperture, color='r', fill=False, ls='-.', label='Comp')
                 ref_circle_sky = plt.Circle((finXRefCent[0], finYRefCent[0]), minAperture+minAnnulus, color='r', fill=False, ls='--', lw=.5)
-            plt.imshow(np.log10(sortedallImageData[0]), origin='lower', cmap='Greys_r', interpolation=None, vmin=np.log10(np.nanmin(sortedallImageData[0][sortedallImageData[0] > 0])), vmax=np.log10(np.nanmax(sortedallImageData[0][sortedallImageData[0] > 0])))  #,vmax=np.nanmax([arrayTargets[0],arrayReferences[0]]))
+
+            med_img = median_filter(sortedallImageData[0],(4,4))
```
rzellem, line 2547: nice addition! thanks!

exotic/exotic.py
```diff
@@ -2670,8 +2758,8 @@ def main():
                 'a2': 0,             #Flux lower bound
             }
 
+        
         phase = (goodTimes-prior['tmid'])/prior['per']
```
rzellem, line 2762: why use this and not the function?

rzellem (changes requested):
Please address comments in-code.

exotic/exotic.py
pearsonkyle: The function returns a phase between -1 and 1. I need the orbital phase with the integer in front of it indicating the orbital epoch. The equation below that line should match your paper equation for predicting transits...

exotic/exotic.py
pearsonkyle: Cus it works

exotic/exotic.py
pearsonkyle: This happens after the images are aligned. The lines below this save the  centroid positions from the PSF fits so they can get plotted later.

exotic/exotic.py
rzellem: Ok...what is the benefit of changing the function from what it was to what you want to make it? Speed?

rzellem: Kyle sez: "faster"

rzellem (approved):

PR #365, opened 2020-10-02
Issue 346: Fixed File Corruption Check

### LINKED ISSUE #346: Fix File Corruption Check

tamimfatahi:
Closes issue #346 . Also, I introduced print(f"{}) as it is a better way for printing (only a few were changed, for now. Waiting for the green-light to do more similar to it).


jpl-jengelke (commented):
I'll approve it. 

One comment is that there is some repeatable code in the `_fail` methods. A `str` could be defined and then passed into a single error method to perform the same.

jpl-jengelke (approved):

Issue #260, opened 2020-08-18
Allowing EXOTIC to run from either directory


mjs2369:

tamimfatahi:
Looks like we'll be strictly running from /EXOTIC only as we don't want to touch internal packages

rzellem:
No one seems to be complaining about this...so let's close it.

PR #366, opened 2020-10-02
Issue 260: Changed working directory to ./EXOTIC

tamimfatahi:
Closes Issue #260

### DIFF
--- a/exotic/exotic.py
+++ b/exotic/exotic.py
@@ -78,6 +78,7 @@ import platform
 import warnings
 import argparse
 import glob as g
+from pathlib import Path
 from io import StringIO
 
 # data processing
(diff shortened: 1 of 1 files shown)


jpl-jengelke:
There are some concerns about this in relation to using EXOTIC as a module, the location of log file, actually knowing the right directory, etc.

jpl-jengelke (changes requested):
It's unclear if this will work. If it is run via the PyPi install will it move up a directory? If it is run via python3 exotic/exotic.py will it move up a directory? How has it been tested? 

Probably a more robust solution for this uses case is to make a wrapper script at the top of the repo named exotic.py that will execute the application one layer down.

PR #335, opened 2020-09-09
Issue 272: Improved Plate Solution API and added in LCO band for filtering.

tamimfatahi:
Added a safeguard for plate solution using tenacity and an extra LCO band as requested by Michael Fitzgerald. Issue #272

PR #375, opened 2020-10-08
#335: add interactive plotting mode for scaling

rzellem:


rzellem:
Need to fix something...brb.

Issue #373, opened 2020-10-07
Move all user inputs to the inits file


rzellem:
Right now, most of the prompts are also moved to the inits file, but the user must still:

1. select `2-1-2`
2. point to their inits file
3. confirm that they want to review the NEA scraped parameters or override them with their own

Rather, it would be nice to move these 3 prompts into the inits.json file as well so that if they are defined (i.e., not `null`), then EXOTIC will not ask the user how they want the code to be run

Then, it would be nice to be able to call EXOTIC like this:
`python exotic/exotic inits.json`
that way EXOTIC will fully run automatically **without** any user interaction at all

tamimfatahi:
Also, add in option of cosmic ray filter.

PR #442, opened 2020-12-02
Issue 213 & 373: Automation of user inputs 

### LINKED ISSUE #213: Improve inits.json file
Suggestions from Kalee Tock:
Also (not critical for this release, but just putting it here so you have it for later): Scott and I met earlier this afternoon and were discussing how it would be awesome if there could be an option where you can tell EXOTIC not to ask you any questions once you start it going.  E.g. do *not* ask for permission to upload to astrometry.net (maybe put that into the inits file if you really need permission?) or whether you want EXOTIC to calculate the limb-darkening parameters, or what filter you are using and what the bandpass is if it is custom (could that be in the inits file too?) or which values you want if the ones in the inits file conflict with what gets scraped from the Nasa Exoplanet Archive.  (Example: it scraped 0.0017 when I had entered 0.00169.  I would love to have an option in the inits file where I can just say: “If there is a conflict, use my values, not the Nasa Exoplanet Archive ones.  I looked them up and the ones in the inits file are the ones I want.“)  Also, if the pixel coordinates found from the plate solution are only a pixel or two different from the ones entered by the user, then it has to be the same star, right?  Is it really necessary to ask the user whether they want to adjust the pixel coordinates, and then do a whole second plate solution?  Also, for the command line entry option, would it be possible to ask one question about the Nasa Exoplanet Archive parameters rather than so many separate ones?  That way, the user could blanket-authorize the scraped values rather than having to examine them one by one, which is especially tedious if you are running the same system multiple times.  (Perhaps if the user says no to the blanket authorization, then the system could ask one by one.)
4:10
(Maybe there could be a “professional version” for people who know things like how to calculate limb-darkening parameters, and an “amateur version” for people who just get confused by all that.  :slightly_smiling_face:  )

tamimfatahi:
Closes #213 & #373

Users can now use the command line to bypass planetary parameters differences if they choose to by entering `override`. Also, prereduced data is now workable with command line arguments. 

All command line arguments:

- `-i inits.json -o reduce override`
- `-i inits.json -o prereduced override`
- `-i inits.json -o reduce`
- `-i inits.json -o prereduced`
- `-i inits.json`

if users use the `-o` argument, they must enter 1 or more parameters. Also, you cannot enter both `reduce` and `prereduced` as that will result in an error. If `override` is not a given parameter, it will assume the user still wants to check between the NEA and their JSON file.

Also, `override` will **not** scrape the NEA. For a possible future update, it might be better to check the inits.json file for missing values and fill them in with the NEA values. However, this does give flexibility to users who don't have internet access at the time of reduction and would rather just use their JSON values.

Users can also type in `-h` and it will show a list of choices for `-o` since it must be spelled correctly.

### DIFF
--- a/exotic/exotic.py
+++ b/exotic/exotic.py
@@ -1829,8 +1829,11 @@ def realTimeReduce(i, target_name):
 
 
 def parse_args():
-    parser = argparse.ArgumentParser(description="Start exotic with an initialization file to bypass all user input.")
-    parser.add_argument('-i', '--init', help="choose an inits.json file", type=str, default='')
+    parser = argparse.ArgumentParser(description="Start exotic with an initialization file to bypass all user inputs.")
+    parser.add_argument('-i', '--init', default='', type=str, help="entering inits.json file", )
+    parser.add_argument('-o', '--options', nargs='+', default='', type=str,
+                        choices=['prereduced', 'reduce', 'override'],
+                        help="added options to inits.json file",)
     return parser.parse_args()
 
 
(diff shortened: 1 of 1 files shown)


rzellem (approved):

MJFFowler:
I was just about to ask whether the use of pre-reduced data could be implemented on the command line ... but you beat me to it! Many thanks @tamimfatahi !

PR #454, opened 2020-12-04
Issue 447: Headers w/ Plate Solutions now usable

### LINKED ISSUE #447: If images already plate solved, then use that solution

tamimfatahi:
Closes #447

When creating a the reference file, the header was auto-generated with a standardized one, not the original. I went ahead and kept track of all the original headers of the files so that when creating the ref_file, it uses the original one after being sorted by time.


tamimfatahi:
One moment. Found a bug in testing.

rzellem (approved):

Issue #403, opened 2020-11-02
Allow EXOTIC to Open Large Files


rzellem:
They aren't really large files. Modern day cameras can produce 60MB-120MB each, so EXOTIC needs to be able to deal with 400 120MB files. So you simply can't hold them all in memory at once. (edited) 
2:21
I'm working with an 11" telescope at a school nearby and they are 120MB an image.
2:21
lots of MPs... but thats the way of the future
2:22
24.2MB is standard LCO size which is a relatively limited 29' by 19' frame and an appropriate pixel scale.
2:24
The solution would be to convert the image arrays into numpy binary files and read them from there (I've pointed this out before)... it would be unnoticeably marginally slower than holding them in memory
2:25
https://towardsdatascience.com/what-is-npy-files-and-why-you-should-use-them-603373c78883

MediumMedium
What is .npy files and why you should use them…
Why you should always save your data as .npy instead of .csv
Reading time
4 min read
Aug 19th (268 kB)
https://miro.medium.com/max/1200/1*vA1XbqldcFwG2EvXWEh2CA.jpeg
2:26
it would also be much more efficient than getting a computer to use its swap memory, which I imagine if you are using anything other than MO data on a relatively standard computer on a normal transit dataset it will do so more than 70% of the time.

Michael Fitzgerald  2:32 PM
DSLR images will be about 60-120MB+ in RAW mode, sometimes much more if it is a high MP camera
2:32
So there is no avoiding it - you can't hold all the images in memory

rzellem:
hi @mfitzasp - any updates on the progress of this issue? can we help out with the coding? thanks!

PR #463, opened 2020-12-11
Decreasing Memory Usage

tamimfatahi:
Here is the memory usage of EXOTIC that is currently on develop: 
Original

Here is the memory usage of EXOTIC after removing the scaling feature, which was scaling from `int16` -> `float32`:
full_reduction_no_32

Both are complete reductions. Also, scaled-down airMassList from `float64` to `float32`.

The big spike you are seeing towards the beginning of the reduction is where we open the FITS files and append them to the `list`. You'll notice a huge drop once we convert those `lists` to `numpy arrays`. Although `list` is great due to how fast it is when appending large amounts of data, it costs a lot of memory. The increase towards the end is due to the nested samplers and how it causes sudden drops after iterations.

It's also interesting to note that time did increase with these additions. Might need multiple testings on complete reductions on that as I closed PyCharm during the second run. However, I did run the beginning portion of the code multiple times leading up to the alignment and the usage was consistent where it would utilize ~600 MB in the original vs. ~400 MB with the scaling taken out.

**Edit:** Also, it seems like there is still memory left towards the end if I'm correct? Shouldn't it dump to 0 MB when done reducing completely and exiting EXOTIC? 

**Edit2:** Went ahead and deleted two of our largest arrays (`allImageData` & `sortedallImageData`) since they hold data from the FITS files right before the nested sampler. You'll notice a sharp drop around ~250 seconds, which is where the deletion occurs.

final_complete_reduction

If we can find a way to open the FITS files properly (issue #403), that would be key. The rest should be up to garbage collection/organization by separating these key portions into their own functions. Right now, I'm simply deleting them from memory and hoping it will free memory space.


jpl-jengelke:
Memory usage may not drop to zero immediately at the end. There are some tools in Pycharm that are loaded into memory for the analysis and runtime, plus module imports. The trick is determining the overhead. If you write a simple app that prints "Hello World" 10k times, the overhead will be clear.

tamimfatahi:
This was done inside the terminal. Would that cause these issues? Interesting about the overhead though. Would adding a garbage collector in the end help?

tamimfatahi:
**Update 2:** I went ahead and overwrote `sortedallImageData`with `allImageData`. It seems though the deletion of the `numpy array ` does not free it from memory space. When `allImageData` was deleted in the prior update, it was a `list` that caused an increase in free memory. 

overwrote allImageData

rzellem:
@tamimfatahi - I would take a look at how we deal with memory when reading in the fits files - I think we use the `del` command, but I would check EXOTIC and see if that's what you're also using or not.

rzellem (approved):

PR #472, opened 2020-12-16
fixed missing header value

pearsonkyle:


exotic/exotic.py
```diff
@@ -1348,9 +1348,12 @@ def image_alignment(imagedata, roi=1):
...
         for i in range(len(hdul[0].header)):
-            if hdul[0].header['COMMENT'][i].split(' ')[0] == 'scale:':
-                imscalen = float(hdul[0].header['COMMENT'][i].split(' ')[1])
-                break
+            try:
+                if hdul[0].header['COMMENT'][i].split(' ')[0] == 'scale:':
+                    imscalen = float(hdul[0].header['COMMENT'][i].split(' ')[1])
+                    break
+            except:
+                imscalen = -1
```
rzellem, line 1356: any reason you don't want to make it a string like "N/A" instead?

rzellem (changes requested):

exotic/exotic.py
pearsonkyle: This variable is required to be a float otherwise it throws more errors below

rzellem (approved):

PR #479, opened 2020-12-21
German translations

LeonBew:
The German documentation files need to be merged into main.


Documentation/German/Beispiel_Output.txt
```diff
@@ -0,0 +1,264 @@
+Importing Python Packages - please wait.
```
rzellem, line 1: hi @LeonBew - why do you also include this file here? Just want to make sure you want it part of this PR.

rzellem (changes requested):

rzellem (approved):

Documentation/German/Beispiel_Output.txt
rzellem: Ah never mind, it is the sample data. Thanks! :)

Issue #485, opened 2020-12-29
Filter for 0 prior to light curve fitting


rzellem:
We filter for Nan's and infinites. The infinite would arise if the comp star is 0. We don't have a filter for zeros at the moment

rzellem:
@pearsonkyle - getting new error:

```
Checking for variability in current comparison star... 
INFO: Query finished. [astroquery.utils.tap.core]
Traceback (most recent call last):
  File "exotic/exotic.py", line 3315, in <module>
    main()
  File "exotic/exotic.py", line 2689, in main
    nanmask = np.isnan(arrayFinalFlux) | np.isnan(arrayNormUnc) | np.isnan(arrayTimes) | np.isnan(arrayAirmass) | arrayFinalFlux > 0 | arrayNormUnc > 0
TypeError: ufunc 'bitwise_or' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

tamimfatahi:
I recommend checking out the latest PR. Hopefully will solve this.

tamimfatahi:
Completed in PR #488

PR #488, opened 2021-01-05
Temporary fix for the allImageData and modification of filtering for 0's.

tamimfatahi:
Should hopefully include the fix for issue #485

### DIFF
--- a/exotic/exotic.py
+++ b/exotic/exotic.py
@@ -2229,8 +2229,11 @@ def main():
             exotic_infoDict['filter'] = user_input("Please enter your filter name from the options at "
                                                    "http://astroutils.astronomy.ohio-state.edu/exofast/limbdark.shtml: ",
                                                    type_=str)
-            exotic_infoDict['notes'] = user_input("Please enter any observing notes (seeing, weather, etc.): ",
-                                                  type_=str)
+            exotic_infoDict['notes'] = user_input("Please enter any observing notes (seeing, weather, etc.)."
+                                                  "If none, leave blank and press enter: ", type_=str)
+
+        if not exotic_infoDict['notes'].replace(' ', ''):
+            exotic_infoDict['notes'] = "na"
 
         if fileorcommandline == 2:
             diff = False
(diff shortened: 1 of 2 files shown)


tamimfatahi:
Added in an additional update to this PR for issue #490

exotic/exotic.py
```diff
@@ -2686,7 +2688,7 @@ def main():
                         arrayAirmass = airMassList[~filtered_data]
 
                         # remove nans
-                        nanmask = np.isnan(arrayFinalFlux) | np.isnan(arrayNormUnc) | np.isnan(arrayTimes) | np.isnan(arrayAirmass) | arrayFinalFlux > 0 | arrayNormUnc > 0
+                        nanmask = np.isnan(arrayFinalFlux) | np.isnan(arrayNormUnc) | np.isnan(arrayTimes) | np.isnan(arrayAirmass) | np.greater(arrayFinalFlux, 0) | np.greater(arrayNormUnc, 0)
```
rzellem, line 2691: the two `np.greater` should actually be:

`(arrayFinalFlux <= 0) | (arrayNormUnc <= 0)`

as it flags values that are less than 0

rzellem (changes requested):

exotic/exotic.py
tamimfatahi: That would make `nanmask` all `False` if the values are above`0` unless I'm mistaken. Aren't we filtering for values less than or equal to `0`, making those ones `False`?

rzellem: the nanmask is True for values that need to be filtered out, so you need to ID which values are <= 0

np.isnan returns True if the value is a nan, and therefore needs to be filtered out

tamimfatahi: Oh gotcha, makes sense. Thought `False` would be the ones filtered out since I've done that before. Makes it much more clear when you mention `np.isnan` into this. Thanks!

rzellem: if you look elsewhere in the code, @pearsonkyle then filters out these bad values via:

`filtered_array = array[~nanmask]`

for the record, I also do it the other way where True == value is kept, but @pearsonkyle likes to dance to the beat of his own bongo :P

rzellem (approved):

PR #492, opened 2021-01-06
Pre-reduced Data Update

tamimfatahi:
Updated the pre-reduced data portion of EXOTIC to fix errors and modified some formatting of code. Users who use this will have to enter the exposure time of their imaging files from the command line whether they use command-line arguments or not. Might be good to add the exposure time to the `optional_info` section of inits.json so that it covers them in the future.


exotic/exotic.py
```diff
@@ -2984,55 +2991,57 @@ def main():
...
-                    formatEntered = True
-            fluxFormat = (fluxFormat.lower()).strip()
-            if fluxFormat != "flux":
-                goodFluxes, goodNormUnc = fluxConvert(goodFluxes, goodNormUnc, fluxFormat)
+                    break
 
-            bjdMidTOld = goodTimes[0]
-            standardDev1 = np.std(goodFluxes)
+            if flux_format != 'flux':
+                goodFluxes, goodNormUnc = fluxConvert(goodFluxes, goodNormUnc, flux_format)
```
tamimfatahi, line 3045: Removed these as it didn't look like we were using them for pre-reduced data. Let me know if this is needed to be placed back in.

rzellem (approved):

PR #545, opened 2021-02-23
Meridian Flip during Image Alignment

tamimfatahi:
Added back in feature of image alignment to handle meridian flips.


exotic/exotic.py
```diff
@@ -2523,19 +2519,20 @@ def main():
...
 
-                if np.isclose((apos[1] ** 2).sum() ** 0.5, 0):
+                if np.isclose((apos[1]**2).sum()**0.5,0) and i != 0:
                     print(fileName, "no alignment")
 
                 # Fit PSF for target star
-                xrot = exotic_UIprevTPX * np.cos(arot[1]) - exotic_UIprevTPY * np.sin(arot[1]) - apos[1][0]
-                yrot = exotic_UIprevTPX * np.sin(arot[1]) + exotic_UIprevTPY * np.cos(arot[1]) - apos[1][1]
-                psf_data["target_align"][i] = [xrot, yrot]
+                if 3.0 <= np.abs(arot[1]) <= 3.3:
```
tamimfatahi, line 2528: Checks to see if the image was rotated 180 degrees or 3.14 radians. Added a tolerance check from 3.0 - 3.3

rzellem: what is the image is rotated more (or less) than this threshold- does it break things?

rzellem (changes requested):

Issue #562, opened 2021-03-05
Query Elevation for any given Lat/Long


rzellem:
@pearsonkyle says something about a Google API

pearsonkyle:
https://developers.google.com/maps/documentation/elevation/overview

Issue #495, opened 2021-01-13
Read lat/long from FITS header


rzellem:
From Kalee:

```
if it would be possible to read the observatory lat / long / elevation from the image headers, that would be great.  I do this in the Colab I wrote for my students because understandiong the role of those parameters is not part of the learning goal for my intro assignment.  It was a little tricky because of image header idiosyncracies (e.g. you have to add the negative sign for MObs images, and some of the lat / long data is in strange units) but I was able to get it working.  It would be awesome if EXOTIC could do this internally.
```

rzellem:
You can see the code for making the inits files from fits headers if you like if you make a copy of the Colab notebook and double click on the 4th cell (the code is hidden unless you double-click): https://colab.research.google.com/drive/1UcDfm3z1WnfdOpRwjCQYwDgK9Wh2cU6x?usp=sharing (edited)

tamimfatahi:
Scrape elevation from FITS header

rzellem:
and if elevation doesn't exist in FITS header, use API to query the interwebs (see issue #562)

tamimfatahi:
Completed in PR #805

PR #569, opened 2021-03-05
Issue 562: Added Elevation API scraper based on Lat/Long

tamimfatahi:
This PR doesn't look for lat/long/elev in the header (issue #495), just solves the elevation API scraper (issue #562).


tamimfatahi:
The process EXOTIC goes for is:

1. If the elevation is provided in inits.json, default to that one.
2. Use open elevation API to try and get elevation based on lat/long.
3. If open elevation API fails after trying multiple times, ask the user for elevation.


Just want to state that I did test this out based on the current lat/long for HAT-P-32 b. The elevation in the inits.json is 2616 m, and the one open elevation scraped was 2467 m.

rzellem (approved):

Issue #605, opened 2021-03-24
Try/except for image/output file saving


rzellem:
To throw user an error when saving output rather than crashing

PR #613, opened 2021-04-05
Issue #605: Adding try/except to output files

tamimfatahi:
Please ignore `inputs.py` for now as I didn't want to add too many changes to this branch (my apologies for combining two issues into one branch). `exotic.py` has not been updated yet to conform with `inputs.py` changes as there are many things needed to be taken out of it, so this push should be harmless to it. Will do that after this PR is approved and tested by the team.

Two new files have been creating: 
- `util.py` : Stores functions that are able to be used across different files and 'random' ones as well.
- `output_files.py` : Creates all the output files. try/except is utilized within `exotic.py` for output file handling.


rzellem (approved):

rzellem:
want to have @pearsonkyle take a look before merging into `develop`

PR #659, opened 2021-05-05
Fix typographical errors in latitude input function

joewie:
The `latitude` function in `inputs.py` asks the user for the longitude. This PR fixes that.


pearsonkyle (approved):
Ah, thanks for catching this!

joewie:
> Ah, thanks for catching this!

I didn't, someone on Slack did. 🙂

PR #715, opened 2021-06-05
Update ecla.py with jit on solveme

Zorba256:
Using EXOTIC 0.43.0 with sample-data the run time for solveme function is reduced from around 400 seconds to around 3 seconds by using Numba JIT. Overall EXOTIC run time is reduced from around 1,400 seconds (23 min) to around 1,000 seconds (17 min). This effect should be more pronounced on longer data-reduction runs where "pre data-reduction" stages in overall EXOTIC process are less significant. 
This has been tested successfully with EXOTIC 0.44.2 and the "jitted" results found to be more or less identical to the "un-jitted" results.

### DIFF
--- a/exotic/api/elca.py
+++ b/exotic/api/elca.py
@@ -43,6 +43,7 @@
 # ########################################################################### #
 
 import copy
+from numba import jit
 import numpy as np
 import matplotlib.pyplot as plt
 
@@ -186,13 +187,13 @@ def time2z(time, ipct, tknot, sma, orbperiod, ecc, tperi=None, epsilon=1e-5):
     z[sft < 0] *= -1e0
     return z, sft
 
-
+@jit(nopython=True)
 def solveme(M, e, eps):
     '''
     G. ROUDIER: Newton Raphson solver for true anomaly
     M is a numpy array
     '''
-    E = np.copy(M)
+    E = M.copy() # numba only allows copy() with no arguments
     for i in np.arange(M.shape[0]):
         while abs(E[i] - e*np.sin(E[i]) - M[i]) > eps:
             num = E[i] - e*np.sin(E[i]) - M[i]


rzellem:
@pearsonkyle - please review these code changes to make sure they're fully compatible with both EXOTIC and CITISENS

pearsonkyle:
@Zorba256 Very interesting! What type of machine are you running this on?

pearsonkyle:
I'd like to test this on a few different operating systems before committing to adding a new dependency to exotic.

Zorba256:
@pearsonkyle Sorry i missed your comments earlier - I don't seem to be getting alerted to GitHub notifications as I would like.

I'm running Exotic itself with these changes to elca.py under Ubuntu 20.04 on a Windows10/WSL2 platform. But as I'm sure you know elca.py can be run in stand-alone mode which I do on a native Windows 10 platform for development of the Numba features. So I'm actually running elca.py on Pycharm Community edition for easy access to the debugger. 

I've also run Exotic with the updated elca.py on a native windows 10 install without issues.

I'd managed to convince myself that Numba was already included in the Exotic build but can now find no evidence of that, so your caution about introducing it at this late stage is understandable.

pearsonkyle:
@rzellem can you please try to import numba on your windows machine?


I've been able to run this pull request after a fresh install of exotic on Windows 10 using this:
```
conda create --name exotic python=3.9
conda activate exotic
conda install numpy scikit-image cython
pip install exotic
```
that suggests we may not have to change the requirements.txt file because Numba is included in one of those dependencies. The code works in the WSL and on a mac so we should be good.

rzellem:
> @rzellem can you please try to import numba on your windows machine?
> 
> I've been able to run this pull request after a fresh install of exotic on Windows 10 using this:
> 
> ```
> conda create --name exotic python=3.9
> conda activate exotic
> conda install numpy scikit-image cython
> pip install exotic
> ```
> 
> that suggests we may not have to change the requirements.txt file because Numba is included in one of those dependencies. The code works in the WSL and on a mac so we should be good.

I can confirm that `import numba` does work on my PC, following the EXOTIC installation instructions on our Github and website.

Zorba256:
That's all good news - I knew I'd previously seen Numba in there somewhere!
I'll be interested to see whether users notice the difference in performance when this goes out.
Zorba256:elca-solveme-jitting branch deleted - I'm working in my local copy of the develop branch now.

PR #716, opened 2021-06-05
Update to plot_image function in plotting.py

R0b0ty:
This is a minor change to the **plot_image** function in **_plotting.py_**, for your review.

When running the default **_Exotic_Notebook.ipynb_** in Jupyter Lab, the following error is observed when reading the initial FITS file:

plot_image_error

The root cause is that the `image_downscaled` variable isn't initialized if the image size (`data.shape[0]`) is less than or equal to 2000 pixels.  The `image_downscaled` variable is called when doing plotting and will be non-existent in this case.

The suggested fix is to append an additional `else` statement.  An alternate proposal is to just use the original image, such as:
`else:
    image_downscaled = data`

... or similar

Also addressed trailing spaces


pearsonkyle:
Thanks for catching that. It looks good

pearsonkyle (approved):
Looks good to me

R0b0ty:
Thanks for your review, Kyle. 👍

PR #717, opened 2021-06-07
v0.45.0: updated install/run

rzellem:
### Features
- Simplified EXOTIC installation and run


jpl-jengelke:
I don't think you need to install Pyvo explicitly. They fixed it this morning. Just wheel, then setuptools then exotic on separate lines.

jpl-jengelke:
I would also do a version check for Python 3.8 and above, but we can add that later along with some other stuff.

jpl-jengelke:
The Windows one won't work. Change the name to `.bat` and replace the first line with `ECHO OFF`

rzellem:
> The Windows one won't work. Change the name to `.bat` and replace the first line with `ECHO OFF`

Done. How's this?

jpl-jengelke (approved):

PR #728, opened 2021-06-09
Patch for exiting on bad Python version

jpl-jengelke:
PLUS code cleanup

This might be PR 1 of 2


jpl-jengelke:
NOTE: Not testable using install script until released.

rzellem (approved):

PR #744, opened 2021-06-16
Issue 532: Updated Centroid Fitter

tamimfatahi:
**Do not merge just yet into develop as there is still an existing bug to be discussed in this PR.**

A new file has been added called `fit_centroid_old.py` indicating older functions used for the centroid fitting routine in the past. The majority of the functions in there can be ignored for now and the main function that should be brought to attention is the `fit_centroid_prev()`. 

For the first 12 images of `HAT-P-32b`, the old centroid fitter works well with the updated aperture photometry routine as well as other functions that go along with it. However, once it hits the 13th image, it seems to lose the star completely. The columns indicate the following:

`0`:  x-position
`1`:   y-position
`2`:  amplitude
`3`:  sigma-x
`4`:  sigma-y
`5`:  rotation (all set to 0 as the older centroid fitter didn't deal with rotation)
`6`:  offset

Screen Shot 2021-06-16 at 3 12 53 AM

If you notice in 13th row in the image above, the amplitude drops to 0 (3rd column) and the target star's coordinates have shifted fairly greatly. I opened up the image to check the coordinates (421, 310) and, unsurprisingly, they do not match the star's coordinates. Neither do the comparison star's coordinates for the same image match either.

I'll comment on what portion of the code where I believe the issue is stemming from.

Other than that, I did increase the range of the box to 25 to see if I can at least get through the aperture photometry routine and it did without errors. Hopefully, with this bug fixed, the older centroid fitter should operate fine with the updated functions.


exotic/fit_centroid_old.py
```diff
@@ -0,0 +1,600 @@
...
+
+    # pars = x,y, a,sigx,sigy, rotate
+    def fcn2min(pars):
+        model = star_psf_prev(xv, yv, *pars)
+        return (data[yv, xv] - model).flatten()  # method for LS
+        # return np.sum( (data[yv,xv]-model)**2 ) # method for minimize
+
+    lo = [pos[0] - box, pos[1] - box, 0, 1, 1, 0]
+    up = [pos[0] + box, pos[1] + box, 100000, 40, 40, np.max(data[yv, xv])]
+    res = least_squares(fcn2min, x0=[*pos, *init], bounds=[lo, up], loss=lossfn, jac='3-point')
```
tamimfatahi, line 70: I'm going to assume it's coming from the least squares method. I've played around with the mesh box and gaussian psf from both the older version and newer, yet both yield the same error.

tamimfatahi:
Closing temporarily for now.

PR #780, opened 2021-06-24
Issue 776 & 777: GUI clarity regarding image alignment & MicroObservatory secondary obs code

### LINKED ISSUE #776: Alignment/Plate Solution Option in GUI
Make it clear for users that they could opt to use their image header's WCS to keep track of target & comp coords instead of doing an alignment in the GUI.

### LINKED ISSUE #777: Automatically add Mobs to secondary observer code
If a dataset comes from MicroObservatory (maybe do an search through the image header?), then add the MicroObservatory observer code to the Secondary Observer Code

tamimfatahi:
Issue #776 : Making it clear that users can opt to use WCS in header to keep track of targets. Please give feedback on output text. Closes #776

Issue #777 : Adding `MOBS` to secondary observer code if user uses MicroObservatory data. Closes #777

### DIFF
--- a/exotic/exotic_gui.py
+++ b/exotic/exotic_gui.py
@@ -293,7 +293,8 @@ def main():
                 i += 1
 
                 window_label = tk.Label(root,
-                                        text="""Please enter your calibration file information\n(Note: each calibration type must be in its OWN, SEPARATE folder)""",
+                                        text="""Please enter your calibration file information
+                                        \n(Note: each calibration type must be in its OWN, SEPARATE folder)""",
                                         # font=("Helvetica 14 bold"),
                                         justify=tk.LEFT,
                                         padx=20)  # .pack()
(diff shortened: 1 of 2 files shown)


exotic/exotic_gui.py
```diff
@@ -437,14 +437,15 @@ def main():
...
+                                                  onvalue=True, offvalue=False)
                 platesolve_check.grid(row=i, column=j, sticky=tk.W, pady=2)
                 i += 1
 
                 alignment = tk.BooleanVar()
-                alignment_check = tk.Checkbutton(root, text='Align my images', variable=alignment, onvalue=1,
-                                                 offvalue=0)
+                alignment_check = tk.Checkbutton(root, text="Align my images\n(Note: Do not check if you'd like to use"
+                                                            "\nWCS in header for tracking targets)",
+                                                 variable=alignment, onvalue=1, offvalue=0, justify=tk.LEFT)
```
tamimfatahi, line 448: What do you think about the output text? Can modify if needed.

rzellem: this might be a little confusing for users- to me, it can read as "oh hey, if we do a WCS solution with EXOTIC, then we don't need to align" - I *think* we want to say is that if someone has already done the WCS plate solution for *all* of their images prior to running EXOTIC, then the alignment is unnecessary...but, don't you check now to see if each image is already plate solved? and if so, it overrides and cancels the alignment?

rzellem (changes requested):

rzellem (approved):

rzellem (approved):

PR #781, opened 2021-06-24
Issue 657: Checking for Valid JSON

### LINKED ISSUE #657: Check inits.json file
A few people are forgetting commas at the end of a line in the inits.json file. We should check that with EXOTIC and update the inits file when needed.

tamimfatahi:
EXOTIC will no longer error out when an invalid format of a JSON is given for an initialization file. For example, when a comma is missing, the following will output and loop again to ask for an initialization file: 

`Error: Invalid JSON. Please reformat JSON based on given suggestion:`
	` - Expecting ',' delimiter: line 32 column 13 (char 2551)`

The errors printed above are coming directly from the JSON module and gives a bit more readability. ~~Closes #657~~

### DIFF
--- a/exotic/inputs.py
+++ b/exotic/inputs.py
@@ -1,7 +1,7 @@
 import logging
 import sys
+import json
 from pathlib import Path
-from json import load
 
 try:
     from util import user_input, dms_to_dd, open_elevation, typecast_check, init_params
@@ -121,7 +121,7 @@ class Inputs:
 
     def comp_params(self, init_file, planet_dict):
         with init_file.open('r') as json_file:
-            data = load(json_file)
+            data = json.load(json_file)
 
         user_info = {
             'images': 'Directory with FITS files', 'save': 'Directory to Save Plots',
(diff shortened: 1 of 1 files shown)


tamimfatahi:
This will partially complete issue #657 since it also states to update the user's file as well. I'll reopen it for now though and we can merge to dev without closing the issue.

exotic/inputs.py
```diff
@@ -1,7 +1,7 @@
 import logging
-import json
 import sys
 from pathlib import Path
+from json import load
```
rzellem, line 4: to nitpick- why do a `from jason import load` rather than `import json` and then later on `json.load(json_file`? The only reason I'm pushing back on this is that there might be another use of the word "load" either today, or perhaps in the future, and calling it this other way might be safer in the long term

rzellem (changes requested):

exotic/inputs.py
tamimfatahi: I can us adding more methods in the future from the JSON module. Restored it to it's previous state.

rzellem (approved):

PR #791, opened 2021-07-07
Issue 783 & 788: Pre-reduced data introduced to GUI and fixed RA & DEC conversion issue

### LINKED ISSUE #783: Add "pre-reduced data" feature back to EXOTIC GUI

tamimfatahi:
Updating the GUI to include the pre-reduced data section of EXOTIC. The`inits.json` file has been updated as well to include these new additions, but all older ones will work as well. Closes #783

### DIFF
--- a/inits.json
+++ b/inits.json
@@ -45,7 +45,7 @@
             "Align Images? (y/n)": "y",
 
             "Target Star X & Y Pixel": [424, 286],
-            "Comparison Star(s) X & Y Pixel": [[465, 183], [512, 263]]
+            "Comparison Star(s) X & Y Pixel": [[465, 183], [512, 263], [], [], [], [], [], [], [], []]
     },
     "planetary_parameters": {
             "Target Star RA": "02:04:10",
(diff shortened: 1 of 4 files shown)


inits.json
```diff
@@ -45,7 +45,7 @@
             "Align Images? (y/n)": "y",
 
             "Target Star X & Y Pixel": [424, 286],
-            "Comparison Star(s) X & Y Pixel": [[465, 183], [512, 263]]
+            "Comparison Star(s) X & Y Pixel": [[465, 183], [512, 263], [], [], [], [], [], [], [], []]
```
tamimfatahi, line 48: What do you think about adding more lists for users to fill in additional comp stars in the `inits.json`? Is it too cluttering? The reason why there's ten in total is because EXOTIC has that limit of 10 comp stars.

rzellem: @tamimfatahi - technically, EXOTIC doesn't have an upper limit on the number of comp stars (I believe that's still the case)

Let's add this into the code for now and see what happens with our users

rzellem (approved):

inits.json
tamimfatahi: When I added in the `inputs.py` file, I added that upper limit just in case. I can remove it if necessary. It's on line 429 if you want to take a gander!

PR #806, opened 2021-07-18
Clean up user input validation

mplough:
This PR cleans up checking of valid values during user input validation.  As a result, it will be easier to check against any number of valid values and integer `0` is now a value that can be validated.


tamimfatahi (commented):
Looks good to me. Thank you for the contribution! @rzellem will give the go ahead to merge to develop once he reviews it as well.

rzellem (approved):

PR #828, opened 2021-08-06
Issue 670: Fixing real time reduction

### LINKED ISSUE #670: Fix Realtime Reduction
Need to make instructions more clear about the purpose of the real time reductions

also, need to fix the error:

```
Enter the Planet Name. Make sure it matches the case sensitive name and spacing used on Exoplanet Archive (https://exoplanetarchive.ipac.caltech.edu/index.html): HAT-P-32 b

Enter the directory path where imaging files are located. (Example using the sample data: sample-data/HatP32Dec202017): sample-data/HatP32Dec202017
Traceback (most recent call last):
  File "exotic/exotic.py", line 2762, in <module>
    main()
  File "exotic/exotic.py", line 1681, in main
    exotic_infoDict = inputs_obj.real_time()
  File "/Users/rzellem/Documents/EXOTIC/exotic/inputs.py", line 83, in real_time
    self.info_dict[key] = self.params[key](self.info_dict[key])
TypeError: target_star_coords() missing 1 required positional argument: 'planet'

```

tamimfatahi:
Real time reduction now works again with EXOTIC due to updating the method we use to find target stars. Closes #670 possibly, as I'm not sure if the document needs to be updated more than it is already to clear up what real time reduction's use is since that is also part of the issue.


rzellem (changes requested):
Please also add back in the option to select real-time reduction in the GUI.

rzellem (approved):

rzellem (approved):

PR #866, opened 2021-09-18
Test for `round_to_2()`

brycemcd:
I've been reading more and more about exoplanets and have looking for a way to get involved. I like this project and hope I can help out a little.

This PR adds some documentation and tests to a function in the `utils` package. If this is desired I can continue to add more tests and more documentation.

I've been a software engineer for a long time but have been a hobbyist astronomer for only a year or so. My day job will prevent me from being a prolific contributor to this project but I am happy to use pockets of time I have to writing tests and making the code more pythonic and idiomatic. I want to support the work you're all doing :)

### DIFF
--- a/exotic/tests/test_utils.py
+++ b/exotic/tests/test_utils.py
@@ -1,0 +1,58 @@
+from exotic.utils import round_to_2
+
+
+class TestRoundToTwo:
+    """tests the round_to_2() function"""
+
+    _ARBITRARY_NUMBER = 4.0
+
+    def test_with_one_arg(self):
+        # One arg may have been passed in
+        result = round_to_2(self._ARBITRARY_NUMBER)
+        assert self._ARBITRARY_NUMBER == result
+
+    def test_second_arg_special_zero_case(self):
+
+        # when second argument passed in is 0.0.
+        result = round_to_2(self._ARBITRARY_NUMBER, 0.0)
+        assert self._ARBITRARY_NUMBER == result
+
+    def test_round_to_two_decimal_places(self):
+
+        float_with_long_fractional_part = 2.34567
+        result = round_to_2(float_with_long_fractional_part, 2)
+        assert 2.35 == result
+
+        result = round_to_2(2.000123456)
+        assert 2.0 == result
+
+        # NOTE: it's kind of weird that the second argument would
+        # just be discarded here. Fix it later once I understand
+        # the code better
+        result = round_to_2(float_with_long_fractional_part, 3)
+        assert 2.35 == result
+
+    def test_small_numbers(self):
+        # the meat of the function. This gets into testing the
+        # -int(floor(log10(abs(y)))) expression
+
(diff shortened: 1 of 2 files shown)


brycemcd:
Chronicling a conversation from #869 to make sure the maintainers understand the approach I have been considering for this PR and the others that mention this one:

Once this PR is merged then some code that sets up testing for this codebase will be merged. I can rebase the WIP MRs on the new `develop` branch and remove the WIP from the PR. That should be quick. Those PRs that are open will then just have code limited to the functions they test and will be easier to review.

exotic/tests/test_utils.py
```diff
@@ -0,0 +1,58 @@
...
+
+        # for numbers that are very small, round to two sig figs
+        result = round_to_2(0.000123456789)
+        assert 0.00012 == result
+
+        # for numbers that are very small, and negative, round to two sig figs and keep the negativity
+        result = round_to_2(-0.000123456789)
+        assert -0.00012 == result
+
+        # not sure if this is a bug?
```
rzellem, line 51: doesn't look like a bug to me- it takes the second number, looks for 2 sig figs (non-0s after the decimal), so in this case that would be 0.00012, and then rounds the first number to that number of sig figs - which should be 2.12300 - but the trailing 0s get dropped by Python

exotic/tests/test_utils.py
```diff
@@ -0,0 +1,58 @@
+from exotic.utils import round_to_2
```
rzellem, line 1: why the new tests folder?

rzellem (changes requested):

exotic/tests/test_utils.py
brycemcd: It's a way to organize tests that keeps the tests 1) close to the code and 2) the implementation code uncluttered.

Documentation for python-conventional ways to organize tests is here: https://docs.pytest.org/en/6.2.x/goodpractices.html#choosing-a-test-layout-import-rules I chose the approach that made the most sense.

exotic/tests/test_utils.py
brycemcd: good to know. I wasn't sure if more precision would be needed. I'll update to remove this comment.

exotic/tests/test_utils.py
brycemcd: There are quite a few open source, popular, python repositories that use this paradigm of testing. Check out [qlib](https://github.com/microsoft/qlib) and it's [top level test directory](https://github.com/microsoft/qlib/tree/main/tests), [sklearn](https://github.com/scikit-learn/scikit-learn/tree/main/sklearn) and [numpy](https://github.com/numpy/numpy/tree/main/numpy) to see how they've implemented their test suites. I used these to guide this and my other testing branches.

I'm happy to chat through other paradigms if you would like. I'm also happy to change paradigms in the future (and backport these) as we get more testing code implemented and a different paradigm reveals itself as advantageous.

rzellem (approved):

Issue #881, opened 2021-10-08
Unit Test Coverage For utils.py


brycemcd:
Changes to `utils.py` are challenging to be made safely and quickly. Unit tests will help lower the barrier to making changes to the functions in this file and provide some additional documentation for new people developing EXOTIC. In order to provide adequate test coverage of the file, the following functions should be tested (arbitrary order):

+ [ ] user_input 
+ [x] init_params #883 
+ [x] typecast_check #886 
+ [x] round_to_2 #866 
+ [x] get_val #882 
+ [x] add_sign #882 
+ [x] process_lat_long #876
+ [x] find #894 
+ [ ] open_elevation

PR #882, opened 2021-10-08
Tests and documents `get_val()` and `add_sign()` functions

brycemcd:
Adds another set of tests for #881 

* In `util` module
* Both functions are called only by other functions in the
  module. Could be refactored a bit

### DIFF
--- a/exotic/utils.py
+++ b/exotic/utils.py
@@ -84,7 +84,6 @@ def round_to_2(*args):
         the original number rounded to two non-zero numbers after the decimal place
     """
 
-
     x = args[0]
     if len(args) == 1:
         y = args[0]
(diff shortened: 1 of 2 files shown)


exotic/utils.py
```diff
@@ -99,6 +98,22 @@ def round_to_2(*args):
...
 def get_val(hdr, ks):
+    """
+    Pluck the value for a certain key from myriad possible known keys
+
+    Parameters
+    ----------
+    hdr : dict
+        a dictionary of observatory data
+    ks : list[str]
+        a list of possible keys? PR FIXME: not sure really
```
brycemcd, line 109: This `get_val` function is only called from the `find` function in the same file. The arguments, `hdr` and `ks` are first passed into the `find` function and then passed into this `get_val` function.

`find` is called from `inputs.py` in the `elevation`, `latitude` and `longitude` functions. (see lines 396, 334 and 364 respectively.) The caller in the `latitude` function is looks like this:

```python
lat = find(hdr, ['LATITUDE', 'LAT', 'SITELAT'])
```
Clearly it's looking for a latitude and I'm guessing that different equipment manufacturers output some known expected values for latitude.

Knowing the expected values of `hdr` and `ks`, and where they originate, would help me to provide better method documentation and help anyone else that works on this later add to this method and make it more precise. Any historical perspective would be greatly appreciated.

brycemcd: @kaleetock this is really helpful to a noob like me!

I was struggling to understand that `hdr` meant "FITS Header" but that totally makes sense now in the context of your explanation.

The PR has been updated based on your comments. I think it will help future developers of this project build up their knowledge of how all this works.

Your explanation gives me a good idea to help make the code more clear. I'll open an issue to explain it more detail. Thanks!

PR #945, opened 2022-01-13
Global fitter for multiple light curves

pearsonkyle:
Fit multiple light curves simultaneously while allowing each light curve to have a set of individual priors and a prior that is shared between all light curves.


exotic/exotic.py
```diff
@@ -1919,6 +1919,8 @@ def main():
                 log_info(f"Optimal Annulus: {np.round(minAnnulus, 2)}")
                 bestaperture = str(abs(np.round(minAperture, 2)))
             log_info("*********************************************\n")
+            
+            import pdb; pdb.set_trace()
```
rzellem, line 1923: you left a pdb in here, dude! :P

rzellem (changes requested):

rzellem (approved):

PR #946, opened 2022-01-18
EXOTIC-GUI: Allowing for cleaner window close events.

Shad0wSeven:
This is a very minor enhancement, on the EXOTIC window created by Tkinter, at least in macOS, the (x) button in the top left corner destroys the Tkinter window, however due to the structure of the way the GUI is programmed, this does not exit the program, and instead erroneously creates the next tk window and so on and so forth, until eventually an exception occurs and the program closes. 

Adding `root.protocol("WM_DELETE_WINDOW", exit)` below each Tkinter initialization allows for users to close the window elegantly without the need for a standard keyboard interrupt.


rzellem (approved):

rzellem:
Thank you for adding this fix!!!

PR #979, opened 2022-04-12
Add extension: .gz

tamimfatahi:
Incorporates the ability to search and open .gz files automatically (image.FITS.gz, MObs file defaults).

### DIFF
--- a/exotic/inputs.py
+++ b/exotic/inputs.py
@@ -189,7 +189,7 @@ class Inputs:
 
 
 def check_imaging_files(directory, img_type):
-    file_extensions = ['.fits', '.fit', '.fts', '.fz', '.fits.gz', 'fit.gz']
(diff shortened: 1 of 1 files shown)


exotic/inputs.py
```diff
@@ -189,7 +189,7 @@ def comp_params(self, init_file, planet_dict):
 
 
 def check_imaging_files(directory, img_type):
-    file_extensions = ['.fits', '.fit', '.fts', '.fz']
+    file_extensions = ['.fits', '.fit', '.fts', '.fz', '.gz']
```
rzellem, line 192: hi @tamimfatahi  - should this line:
file_extensions = ['.fits', '.fit', '.fts', '.fz', '.gz']
be instead:
file_extensions = ['.fits', '.fit', '.fts', '.fz', 'fits.gz', 'fit.gz']
otherwise it might accept just any gzipped file…?

rzellem (changes requested):
see comment :)

PR #1009, opened 2022-07-12
Issue 1004: Omega added to inputs 

### LINKED ISSUE #962: HD 80606 argument of periastron not populating from NEA
A user (Subham Samantaray) just ran HD 80606b with EXOTIC and had the code adopt all NEA values automatically. However, looking at the inits file, it did not write the argument of periastron to the file and looking at the data, it looks like EXOTIC thought that this value == 0, instead of its actual value.

Wonder if argument of periastron isn't properly being treated through all parts of EXOTIC.

### LINKED ISSUE #974: Need to ask for omega if missing
Looks like we forgot to have EXOTIC ask the user for argument of periastron (omega) if it is missing from an inits file.

### LINKED ISSUE #1004: add omega to command line
Have to add periastron to command line in order for EXOTIC to not crash :-)

ihuckabee:
Added omega term and prompted user if omega was null :-) 
closes #1004 and closes #974 and closes #962

### DIFF
--- a/exotic/exotic.py
+++ b/exotic/exotic.py
@@ -346,6 +346,7 @@ def get_planetary_parameters(candplanetbool, userpdict, pdict=None):
                      "Ratio of Distance to Stellar Radius (a/Rs) Uncertainty",
                      "Orbital Inclination (deg)",
                      "Orbital Inclination (deg) Uncertainty",
+                     "Argument of Periastron (deg)",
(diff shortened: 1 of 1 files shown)


tamimfatahi (approved):
Looks good to me. Congratulations on your first PR! :^) Waiting on Rob

rzellem (approved):

Issue #1058, opened 2022-10-14
Error when no comparison stars are scraped from AAVSO + CV --> V 


tamimfatahi:
- Fix error when there are no comparison stars to calculate a star's magnitude from AAVSO
- When users use the CV filter, default it to V when finding comparison stars on AAVSO

tamimfatahi:
Completed in PR #1059

PR #1059, opened 2022-10-14
Issue 1058: Fixing variable star errors and filter query for VSP

tamimfatahi:
- EXOTIC should no longer error out when no comparison stars were gathered from AAVSO's VSP API
- When users utilize the CV filter, it will get converted to V for the AAVSO query on finding mags of comp stars from VSP
- When querying VSP, the image scale is now calculated rather than having a hardcoded value
- Code cleanup, ex: numba entirely from code as it isn't being utilized for cleanup, small QOL changes

Closes issue #1058

### DIFF
--- a/exotic/api/elca.py
+++ b/exotic/api/elca.py
@@ -42,7 +42,6 @@
 # ########################################################################### #
 
 import copy
-# from numba import njit
 import numpy as np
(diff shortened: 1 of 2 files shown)


pearsonkyle:
merge this first

rzellem (approved):

Issue #949, opened 2022-01-20
Automatically constrain Tmid


rzellem:
Important if the uncertainty on Tmid >= the length of the orbital period

tamimfatahi:
Seems like this issue is solved (not sure in which PR, tagging @rzellem to verify):

`        
upper = pDict['midT'] + 35 * pDict['midTUnc'] + np.floor(phase).max() * (pDict['pPer'] + 35 * pDict['pPerUnc'])
lower = pDict['midT'] - 35 * pDict['midTUnc'] + np.floor(phase).max() * (pDict['pPer'] - 35 * pDict['pPerUnc'])

# clip bounds so they're within 1 orbit
if upper > prior['tmid'] + 0.5*prior['per']:
    upper = prior['tmid'] + 0.5*prior['per']
if lower < prior['tmid'] - 0.5*prior['per']:
    lower = prior['tmid'] - 0.5*prior['per']
`

tamimfatahi:
Actually, we can add it within the `fit_lightcurve()` as well. Will open a PR and leave it up for review.

tamimfatahi:
Completed in PR #1063

PR #1063, opened 2022-10-14
Issue 949: Mid-Transit Uncertainty limited by Orbital Period

tamimfatahi:
Although it exists within the final fit ([here](https://github.com/rzellem/EXOTIC/blob/main/exotic/exotic.py#L2347-L2354)), I also added it within `fit_lightcurve()` when finding the optimal aperture/annulus size. Let me know if this change is necessary. 

Closes issue #949 nonetheless.


rzellem:
@pearsonkyle - can you please review this one? you know better than I do about this clipping

I think I originally made this issue because a user was finding that EXOTIC was latching on to the incorrect ephemeris due to their Tmid being so large...but I honestly forget.

pearsonkyle (approved):
Looks good, thanks for making those changes to 0.25

PR #1112, opened 2022-12-01
Update exotic.py

chandrunarayan:
Converting i to int before indexing good_jd_times[] array

### DIFF
--- a/exotic/exotic.py
+++ b/exotic/exotic.py
@@ -2210,10 +2210,10 @@ def main():
                 goodTimes = nonBJDTimes
 
                 index_list = [i for i, time_ in enumerate(times) if time_ in nonBJDTimes]
-                good_jd_times = [jd_times[i] for i in index_list]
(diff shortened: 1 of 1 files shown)


chandrunarayan:
The index for the jd_times[] array is sometimes calculated as a floating point number causing a run time error.  I addressed this by doing a (int) cast.  I has worked for my testing.

rzellem:
hi @chandrunarayan - sorry, finally getting around to this- I'm just wondering why the force conversion of i to an int? I thought `enumerate` returned an int? Or is this just for best practices in coding? Thanks!

mikeprimm:
I can confirm your suspicion, @rzellem : adding some prints using 'type':

type(index_list)=<class 'list'>, type(index_list[0])=<class 'int'>

So, I think I can confirm that this change is harmless, but would not result in a net behavior or performance change.

rzellem (approved):

PR #1123, opened 2022-12-12
importing bokeh.io directly from colab.py

alienlifeform:
This may fix the issue with bokeh's TextLike errors


alienlifeform:
Rob, you can delete/reject this untested PR. It was a possible solution to the bokeh version issues.... I was about to test when Kyle found an alternate fix.

PR #1134, opened 2023-01-18
Minor update to support Win installs

jpl-jengelke:


pearsonkyle (approved):
I approve for now. I was able to get exotic and the gui to run after a fresh install of python3.8 on windows. I also tried a virtual environment with python3.7, which failed due to a dependency check upon running the gui, everything installed properly though. I also tried with python3.9 and that failed due to there being too new a numpy version. All tests were conducted in a new conda environment. For now, people should stick to python3.8 on Windows

PR #1136, opened 2023-01-24
Handle dropped/bad frames cleanly

mikeprimm:
This PR attempts to deal with some issues that can occur in less-than-ideal data sets.  Specifically,

- Cases where a frame fails to solve properly
- Cases where a frame solves properly, but turns out to have the centroid of the target correspond to an 'out of frame' location (that is, failure to have solvable centroid)
- Cases where a companion star turns out to be out of frame, resulting in centroid being unsolvable

In all cases, the main goal is to assure that the various generated per-frame data correctly have their per-frame indexes properly adjusted - since there are extensive assumptions that the per-frame index arrays are logically consistent, and of consistent length.  Specifically,

- aper_data[X][i], where [i] is frame index
- psf_data[X][i], where [i] is frame index
- airmass[i], where [i] is frame index
- times[i], where [i] is frame index

This alignment was being done partially, but not completely, and this resulted in various processing problems during model fitting.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -129,3 +129,5 @@ eaConf.json
 get-pip.py
 newfits.fits
 eaConf.json
+exotic.log.*
+pl_names.json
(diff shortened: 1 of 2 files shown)


mikeprimm:
Shifting to new PR, based off of develop vs main branch

PR #1138, opened 2023-01-28
Fix ./run_exotic_linux.sh to handle Ubuntu not putting ~/.local/bin on PATH

mikeprimm:
This is a check and fallback to handle the fact that, oddly, Ubuntu installations do not put the ~/.local/bin directory (where command_scripts get built and installed, by default) on the PATH - more correctly, if you are running a non-login terminal on Ubuntu (e.g. a terminal window on a graphical shell login), it is not on the path; but if a bash is used for a login shell (e.g. ssh into the system), it is (because it is added by .profile, which is only used for login shells...).

In any case, this uses the bash built-in command, hash, to check if exotic-gui is on the path (nominal path), and runs as before if that is true; else, it checks for existence of the command in ~/.local/bin and runs it from there, if so.

### DIFF
--- a/run_exotic_linux.sh
+++ b/run_exotic_linux.sh
@@ -128,12 +128,17 @@ ${pip_runner} install --upgrade exotic
 echo "INFO: Launching EXOTIC user interface. ..."
 if ${pip_runner} freeze | grep -iq 'exotic' ;
 then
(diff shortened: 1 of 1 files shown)


mikeprimm:
Shifting to new PR, based off of develop vs main

PR #1139, opened 2023-01-29
Avoid code duplication

marcocesari:
Since "or" is a short-circuit operator (https://docs.python.org/3.9/library/stdtypes.html#boolean-operations-and-or-not), the condition can be simplified to avoid code duplication.

### DIFF
--- a/exotic/api/nea.py
+++ b/exotic/api/nea.py
@@ -307,9 +307,7 @@ class NASAExoplanetArchive:
                 rprserr = ((rperr / rs) ** 2 + (-rp * rserr / rs ** 2) ** 2) ** 0.5
                 rprs = rp / rs
 
-        if data['pl_ratdor'] is None:
-            data['pl_ratdor'] = pow((data['pl_orbper'] / 365) ** 2, 1 / 3) / (data['st_rad'] * R_SUN.to('au')).value
-        elif data['pl_ratdor'] < 1 or np.isnan(data['pl_ratdor']):
+        if data['pl_ratdor'] is None or data['pl_ratdor'] < 1 or np.isnan(data['pl_ratdor']):
             data['pl_ratdor'] = pow((data['pl_orbper'] / 365) ** 2, 1 / 3) / (data['st_rad'] * R_SUN.to('au')).value
         else:
             print("WARNING: a/Rs can not be estimated from Nasa Exoplanet Archive. Please use an inits file instead.")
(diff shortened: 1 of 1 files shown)


exotic/api/nea.py
```diff
@@ -307,9 +307,7 @@ def _get_params(self, data):
                 rprserr = ((rperr / rs) ** 2 + (-rp * rserr / rs ** 2) ** 2) ** 0.5
                 rprs = rp / rs
 
-        if data['pl_ratdor'] is None:
-            data['pl_ratdor'] = pow((data['pl_orbper'] / 365) ** 2, 1 / 3) / (data['st_rad'] * R_SUN.to('au')).value
-        elif data['pl_ratdor'] < 1 or np.isnan(data['pl_ratdor']):
+        if data['pl_ratdor'] is None or data['pl_ratdor'] < 1 or np.isnan(data['pl_ratdor']):
```
jpl-jengelke, line 310:
```
if data['pl_ratdor'] is None or np.isnan(data['pl_ratdor']) or data['pl_ratdor'] < 1.:
    data['pl_ratdor'] = pow((data['pl_orbper'] / 365.) ** 2, 1. / 3.) / (data['st_rad'] * R_SUN.to('au')).value
```
- Check if none, then if `isnan`, then since it passes those the number can actually be compared.
- Added decimal values in reassignment to guarantee float calculations are respected and not inadvertently rounded.

jpl-jengelke (changes requested):

exotic/api/nea.py
pearsonkyle: I'd be a little careful with the chain of events here.. If `data['pl_ratdor']` is indeed None, which can happen (since NEA doesn't have a complete set of parameters for every target), python will throw an error if you try to compare None to being `<1`, which is why I arranged that condition to be first in this sequence. Otherwise, you can't string it all together and combine it with `or`

rzellem: @pearsonkyle - given this, should we just close this PR and not pull it into dev?

marcocesari: I updated the PR as requested by @jpl-jengelke. Even if `data['pl_ratdor']` is None, it will never be compared to 1, since `or` is a short-circuit operator. Even if it did, it would not raise an exception, since `None < 1 == True` in Python.

pearsonkyle:
Looks good to me, I was able to validate it with the code below. My only worry is that `np.isnan` doesn't like None values but the short circuit `or` works

```
value = None
import numpy as np
if value is None or np.isnan(value) or value < 1:
    print('dude')
```

rzellem:
@jpl-jengelke - can you please review this PR as well, since you had requested changes, and let me know if it should be approved or not? Thanks!

rzellem:
@jpl-jengelke - can you please review this PR and let me know if I can approve it or not?

jpl-jengelke (approved):

PR #1140, opened 2023-01-29
Make random seed settable to make model solving run-to-run consistent

mikeprimm:
The model fitting code, which makes appropriate use of 'Monte Carlo' techniques when producing solutions, has one problem (which can be a concern for both testing and for research purposes): the random number generator is not seeded in a controlled fashion, which results in the pseudorandom number sequence used being inconsistent from run to run, even for unchanged data and code.  On data sets I've worked with, this does tend to result in minor (but not insignificant) changes from run to run.

Tapping into techniques I've seen on other data processing systems, the most simple solution to this is to have the seed for the random number generator be either explicitly controlled, or deterministically generated from the some consistent elements of the source data.  To this end, this PR adds a 'Random seed' optional attribute to the 'options' section of inits.json (for the first case), while generating a high entropy but repeatable seed (based on a SHA-256 hash of the "Planet Name" and the "Published Mid-Transit Time (BJD-UTC)" fields).

Based on your thoughts on this, we could adjust things to make this more optional (that is, only pin the random seed when the 'Random seed' field is set, with no generated version), but I suspect repeated run consistency is probably a good default behavior (IMHO).


mikeprimm:
Switching to new PR based on develop branch

PR #1143, opened 2023-02-01
Make random seed settable to make model solving run-to-run consistent

mikeprimm:
The model fitting code, which makes appropriate use of 'Monte Carlo' techniques when producing solutions, has one problem (which can be a concern for both testing and for research purposes): the random number generator is not seeded in a controlled fashion, which results in the pseudorandom number sequence used being inconsistent from run to run, even for unchanged data and code. On data sets I've worked with, this does tend to result in minor (but not insignificant) changes from run to run.

Tapping into techniques I've seen on other data processing systems, the most simple solution to this is to have the seed for the random number generator be either explicitly controlled, or deterministically generated from the some consistent elements of the source data. To this end, this PR adds a 'Random seed' optional attribute to the 'options' section of inits.json (for the first case), while generating a high entropy but repeatable seed (based on a SHA-256 hash of the "Planet Name" and the "Published Mid-Transit Time (BJD-UTC)" fields).

Based on your thoughts on this, we could adjust things to make this more optional (that is, only pin the random seed when the 'Random seed' field is set, with no generated version), but I suspect repeated run consistency is probably a good default behavior (IMHO).

### DIFF
--- a/exotic/exotic.py
+++ b/exotic/exotic.py
@@ -1686,6 +1686,10 @@ def main():
         else:
             pDict = userpDict
             CandidatePlanetBool = False
+        # Seed random number generator (for run to run consistency)
+        if exotic_infoDict['random_seed']:
+            log_info(f"Setting random number seed to {exotic_infoDict['random_seed']}")
+            np.random.seed(exotic_infoDict['random_seed'])
 
         if fitsortext == 1:
             # Only do the dark correction if user selects this option
--- a/exotic/exotic.py
+++ b/exotic/exotic.py
@@ -57,6 +57,7 @@ warnings.simplefilter('ignore', category=AstropyDeprecationWarning)
 
 # standard imports
 import argparse
+import hashlib
 # Image alignment import
 import astroalign as aa
 aa.PIXEL_TOL = 1
@@ -1689,7 +1690,10 @@ def main():
         # Seed random number generator (for run to run consistency)
         if exotic_infoDict['random_seed']:
             log_info(f"Setting random number seed to {exotic_infoDict['random_seed']}")
-            np.random.seed(exotic_infoDict['random_seed'])
+        else:
+            exotic_infoDict['random_seed'] = int.from_bytes(hashlib.sha256(f"{pDict['pName']}:{pDict['midT']}".encode()).digest()[0:4], byteorder='little')
+            log_info(f"Generated random number seed {exotic_infoDict['random_seed']}")
+        np.random.seed(exotic_infoDict['random_seed'])
 
         if fitsortext == 1:
             # Only do the dark correction if user selects this option
--- a/exotic/inputs.py
+++ b/exotic/inputs.py
@@ -36,7 +36,8 @@ class Inputs:
             'elev': None, 'camera': None, 'pixel_bin': None, 'filter': None, 'notes': None,
             'plate_opt': None, 'aavso_comp': None, 'tar_coords': None, 'comp_stars': None,
             'prered_file': None, 'file_units': None, 'file_time': None, 'phot_comp_star': None,
-            'wl_min': None, 'wl_max': None, 'pixel_scale': None, 'exposure': None
+            'wl_min': None, 'wl_max': None, 'pixel_scale': None, 'exposure': None,
+            'random_seed': None
         }
         self.params = {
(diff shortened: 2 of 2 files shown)


exotic/exotic.py
```diff
@@ -1686,6 +1687,13 @@ def main():
         else:
             pDict = userpDict
             CandidatePlanetBool = False
+        # Seed random number generator (for run to run consistency)
+        if exotic_infoDict['random_seed']:
+            log_info(f"Setting random number seed to {exotic_infoDict['random_seed']}")
+        else:
+            exotic_infoDict['random_seed'] = int.from_bytes(hashlib.sha256(f"{pDict['pName']}:{pDict['midT']}".encode()).digest()[0:4], byteorder='little')
+            log_info(f"Generated random number seed {exotic_infoDict['random_seed']}")
+        np.random.seed(exotic_infoDict['random_seed'])
```
jpl-jengelke, line 1696: How to seed the RNG really depends on the underlying algorithm being used. In the case of Numpy, it is PCG64, which requires a 128-bit initial seed but also a 128-bit internal state vector. In this case a 256-bit hash would be sufficient since the `seed` argument is used to generate both values. (Over-/under-seeding is not recommended -- may waste CPU and actually have unintended consequences on the output stream. Just mentioned here -- it's not currently the case.) 

However, to guarantee a specific RNG, it should be [specifically declared when invoking Numpy](https://numpy.org/doc/stable/reference/random/generator.html). Scrolling down the page near the bottom should be code for declaring the RNG. **Note:** I believe the PCG64DXSM has a more even statistical distribution of generated values, and [it is **recommended** for any large numerical sampling algorithm](https://numpy.org/doc/stable/reference/random/upgrading-pcg64.html). That being said, I have yet to see results of any of Numpy's implementations being tested against the [gold-standard DIEHARDER suite](https://webhome.phy.duke.edu/~rgb/General/dieharder.php).

I have not deep-dived on this so I cannot provide specifics on seeding PCG64DXSM, but if statistical regularity is a concern, this appears the way to go.

exotic/exotic.py
```diff
@@ -1686,6 +1687,13 @@ def main():
         else:
             pDict = userpDict
             CandidatePlanetBool = False
+        # Seed random number generator (for run to run consistency)
+        if exotic_infoDict['random_seed']:
+            log_info(f"Setting random number seed to {exotic_infoDict['random_seed']}")
+        else:
+            exotic_infoDict['random_seed'] = int.from_bytes(hashlib.sha256(f"{pDict['pName']}:{pDict['midT']}".encode()).digest()[0:4], byteorder='little')
```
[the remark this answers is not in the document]
mikeprimm-mirific: Definitely no problem with displaying the fixed seed - I was kind of on the fence on whether to add the 'noise' or not.  I did want to display the non-fixed seed because it does make the value used part of the logs - allowing for the value to be opted to be fixed later, for future runs, for example.

The point IS ultimately to enable run to run consistency - which is generally viewed as a positive thing when it comes to research output artifacts: without it, sharing data, code and methodology is insufficient to assure reproducibility, both by ourselves and by others.   For any given run, a deterministic seed is used - whether provided by us systemically, or via the seed initialization in the numpy runtime (which is neither deterministic nor repeatable, as current time and other entropy sources are generally used for such things) - the result is only the randomness of the selection of sequence used, not the quality of the randomness within that sequence: if the quality of the RNG sequence algorithm is poor, using a fixed or uncontrolled seed will not fix it.

exotic/exotic.py
mikeprimm-mirific: Right - logically, there are two things: the quality of the RNG (which is quite good for PCG64DXSM), and the seed used to drive the resulting deterministic (whether or not it is good or bad - the bits out of the RNG are a deterministic sequence, as defined by the algorithm and the seed.  We're using the same RNG run to run, but the method by which the seed is determined, when not provided, is not controlled nor recorded - when uninitialized, the seed is synthesized (generally from clock time and other entropy sources) ONCE, and the sequence generated from that is determined; when initialized, the same thing happens - the quality of the randomness is a function of the algorithm, not the seed.  No seed is good or bad, given a good RNG - it's only what drives the resulting sequence.

Main point is that an uncontrolled initial seed doesn't make your random sequence any higher quality - it only makes it harder to reproduce from run to run: the entropy of the sequence within the run is purely a function of the algorithm, not the seed.

PR #1145, opened 2023-02-01
Fix ./run_exotic_linux.sh to handle Ubuntu not putting ~/.local/bin on PATH

mikeprimm:
This is a check and fallback to handle the fact that, oddly, Ubuntu installations do not put the ~/.local/bin directory (where command_scripts get built and installed, by default) on the PATH - more correctly, if you are running a non-login terminal on Ubuntu (e.g. a terminal window on a graphical shell login), it is not on the path; but if a bash is used for a login shell (e.g. ssh into the system), it is (because it is added by .profile, which is only used for login shells...).

In any case, this uses the bash built-in command, hash, to check if exotic-gui is on the path (nominal path), and runs as before if that is true; else, it checks for existence of the command in ~/.local/bin and runs it from there, if so.

### DIFF
--- a/run_exotic_linux.sh
+++ b/run_exotic_linux.sh
@@ -128,12 +128,17 @@ ${pip_runner} install --upgrade exotic
 echo "INFO: Launching EXOTIC user interface. ..."
 if ${pip_runner} freeze | grep -iq 'exotic' ;
 then
-    # Ubuntu doesn't put this on path...
-    if [ -x ~/.local/bin/exotic-gui ] ;
+    if hash exotic-gui > /dev/null ;
     then
-        bash -c "~/.local/bin/exotic-gui"
-    else
         bash -c "exotic-gui"
+    # Ubuntu doesn't put this on path...
+    else
+        if [ -x ~/.local/bin/exotic-gui ] ;
+        then
+            bash -c "~/.local/bin/exotic-gui"
+        else
+            echo "ERROR: cannot find exotic-gui"
+        fi
     fi
 else
     echo "ERROR: Unable to launch EXOTIC, installation failed. Please verify installation"
--- a/run_exotic_linux.sh
+++ b/run_exotic_linux.sh
@@ -128,7 +128,13 @@ ${pip_runner} install --upgrade exotic
 echo "INFO: Launching EXOTIC user interface. ..."
 if ${pip_runner} freeze | grep -iq 'exotic' ;
 then
-    bash -c "exotic-gui"
+    # Ubuntu doesn't put this on path...
+    if [ -x ~/.local/bin/exotic-gui ] ;
+    then
+        bash -c "~/.local/bin/exotic-gui"
+    else
+        bash -c "exotic-gui"
+    fi
 else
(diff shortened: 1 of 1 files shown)


jpl-jengelke:
This is indeed an issue, but not for the reason suggested. In many deployments, the base Python tooling is only available to privileged users. When installations happen in that case, the Unix Systemd standard suggests that tooling be installed into `~/.local', and so Python3 `pip` is abiding by the standard. Ubuntu adopted `systemd` in approximately the 2013 timeframe.

https://unix.stackexchange.com/a/407029 

The installation of `EXOTIC` in the user space (system Py3 modules are confounding when users need specified ones) necessitates adjustments: 
- Ensure the local pip is current:
`python3 -m pip install --upgrade --force-reinstall --no-cache-dir pip`
- Ensure local relies on current modules: 
`pip3 install --force-reinstall --no-cache-dir -r requirements.txt`
- Export your PATH to include the local user install directory: (some variation of) `export PATH=~/.local/bin:$PATH`
  - Note that this should almost universally be performed on a `systemd` box so that not only versions of end-point applications are referenced, but all their dependencies should be referenced appropriately, also. (Installs likely won't update the shared system versions of applications, so there will be unanticipated dependency mismatches and/or errors if this isn't the case.)

Then `exotic-gui` will run as expected.

jpl-jengelke (changes requested):
Please see inline comments. Preference for now is to autodetect the existence of local installs and use those.

run_exotic_linux.sh
```diff
@@ -128,7 +128,18 @@ ${pip_runner} install --upgrade exotic
 echo "INFO: Launching EXOTIC user interface. ..."
 if ${pip_runner} freeze | grep -iq 'exotic' ;
 then
-    bash -c "exotic-gui"
+    if hash exotic-gui > /dev/null ;
```
jpl-jengelke, line 131: Instead of testing for `exotic-gui`, preference is to test for `~/.local/bin/exotic-gui` then add it to the PATH. 
Pseudocode: 
> if local-exotic-gui-in-bin ; then
>     export PATH=local-bin:${PATH}
> fi

mikeprimm: If you prefer, but I'll say why I don't think it is preferable:
- If exotic-gui is on the path, assuming that the ~/.local/bin version is preferable defeats PATH intention: it would cause that version to be run despite there being an intended version on the PATH
- Adding to the PATH has implications for any subprocesses or other commands executed later in the script - this is contrary to the intention of just dealing with the command in question.

On the point of the 'correct' way to get things right - the concern is the default environment as it exists on a major distribution, per default installation: ~/.local/bin is in .profile, but .profile is only applicable to a login shell, and x terminals aren't login shells, despite being the most common place to run a Tk app like exotic-gui.  Prescribing how to get the environment right kind of seems contradictory to what the ./run-exotic-linux.sh script does, which seems to be to assure that everything that is in place and needed to run is in fact in place - if that only works given some additional configuration modification, I'm not sure what purpose it serves.

PR #1151, opened 2023-02-07
Add support for demosaic, for sake of color scopes/cameras (e.g. Unistellar)

mikeprimm:
This PR adds two new optional settings to inits.json:

- "Demosaic format" (dictionary symbol demosaic_fmt) - if defined and non-null, this specifies the Bayer pattern of the images in the FITS files.  Valid values include 'RGGB', 'BGGR', 'GRBG', 'GBRG'.  RGGB is likely most common pattern for color imagers.  Demosaicing is done using a bilinear interpolation - I opted for this, not because it is the best method, but because I suspected that its simple linear interpolation was less likely to result in added luminocity 'noise' from things like edge detection that visually superior methods use.
- "Demosaic output" (dictionary symbol demosaic_out) - this specifies the disposition of the demosaiced data, since we don't leverage having 3 primaries.   Values include the following:
   - [ weightR, weightG, weightB ]: this allows a custom camera-specific set of weights to be applied, allowing the output to be determined by R * weightRed + G * weightG + B * weightB.  Weights are normalized before use, such that weightR+weightG+weightB = 1.0.
   - 'gray' : converts the RGB to a gray scale, using standard luminosity method (Y = 0.299 * R + 0.587 * G + 0.114 * B) as implemented by rgb2gray function in skimage library.   Equivalent to [ 0.299, 0.587, 0.144 ].
   - 'red', 'green', 'blue': this pulls just the corresponding channel.  This may be worth considering for, say, chromaticity testing (but I need to investigate some false positive targets to see if this actually works in practice).   Equivalent to [ 1, 0, 0 ], [ 0, 1, 0 ], and [ 0, 0, 1 ], respectively.
   - 'blueblock': this is a variant of 'gray', except it is done after zeroing the blue color channel.  Cannot say how effective this is, but it seems an interesting experiment.  Equivalent to [ 0.299, 0.587, 0.0 ].
   
If not specified, 'green' is used,  corresponding to the AAVSO recommendations I've seen for DSLR and color CMOS.  This avoids the issue of relative color weight variability, while leveraging the 50% of the pixels that are green - and the curve for green on most color cameras is often similar to a V filter, although CG is probably most appropriate.

The processing results in an identically sized and typed imageArray, so as to minimize possibilities of other downstream implications.
 
 The interactive CLI options for setting the fields is included here, but not currently with exotic_gui.py support (I was concerned about window size and layout...).

### DIFF
--- a/exotic/exotic.py
+++ b/exotic/exotic.py
@@ -96,9 +96,11 @@ from scipy.signal import savgol_filter
 from scipy.ndimage import binary_erosion
 from skimage.util import view_as_windows
 from skimage.transform import SimilarityTransform
+from skimage.color import rgb2gray
 # error handling for scraper
 from tenacity import retry, stop_after_delay
-
+# color, color_demosaicing
+from colour_demosaicing import demosaicing_CFA_Bayer_bilinear
 # ########## EXOTIC imports ##########
 try:  # light curve numerics
     from .api.elca import lc_fitter, binner, transit, get_phase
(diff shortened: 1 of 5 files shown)


rzellem:
Thanks, Mike! That’ll make my life much easier when I do pull this PR.

We did test your other PRs and they seem to working great on develop. We will likely wait for this PR before doing a release.


Dr. Rob Zellem, PhD, MSc (he/him)
Exoplanet Astronomer
Exoplanet Discovery and Science (3262)

JPL   |   jpl.nasa.gov
4800 Oak Grove Dr, Mail Stop 169-320
Pasadena, California 91109
M 626-379-9418

________________________________
From: mikeprimm ***@***.***>
Sent: Thursday, March 23, 2023 11:36:47 AM
To: rzellem/EXOTIC ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [EXTERNAL] Re: [rzellem/EXOTIC] Add support for demosaic, for sake of color scopes/cameras (e.g. Unistellar) (PR #1151)


Actually, I see other PRs were pulled - I'll rebase, retest, and do the comments tonight :)

mikeprimm:
OK - rebase is done - I'll kick off a couple of test runs to check sanity

mikeprimm:
Comments for inits.json should be in (seems to live in a couple places - I think I've found them, though).  Clean test is running now, but should be past where it'd have broke, if something was off

mikeprimm:
OK - done with testing here - let me know if there are any other changes needed before merging :)

mikeprimm:
Cool - let me know if I can help on anything else with the release!

mikeprimm:
Rebased to be current with present state of 'develop' branch

exotic/exotic.py
```diff
@@ -1938,6 +1978,8 @@ def main():
 
                 # CALS
                 imageData = apply_cals(imageData, generalDark, generalBias, generalFlat, i)
+                # Demosaic, if needed
```
jpl-jengelke, line 1981: How does it know "if needed"?

jpl-jengelke (commented):
Lean to approve this, based on answers to comments.

exotic/exotic.py
rzellem: @mikeprimm - are you missing an if statement here to trigger if the data needs to be demosaic-ed?

rzellem (changes requested):

exotic/exotic.py
mikeprimm-mirific: no - like apply_cals just before it, the function implementation just responds to the fact that the input parameters are None when this particular type of calibration is being called for.

rzellem (approved):

PR #1153, opened 2023-02-14
remove barycorrpy to investigate 404 issue

alienlifeform:


alienlifeform:
I will test in my account, but this PR is in place in case we need it

rzellem:
Barycorrpy was fixed. PR no longer needed.

PR #1163, opened 2023-03-25
Shift to 'with' for file - avoid error in finally path when bad

mikeprimm:
The current corruption_check has a problem, due to the hdu1 being an undefined object when an exception is thrown, which results in an exception during the finally path (getattr cannot be called on an undefined object)).

The change switches to 'best practices' for a file like object, using a 'with' to bound the open and cause the close/cleanup to be automatically completed, consisten with fits.open guidance here - https://docs.astropy.org/en/stable/io/fits/index.html

### DIFF
--- a/exotic/exotic.py
+++ b/exotic/exotic.py
@@ -565,14 +565,10 @@ def corruption_check(files):
     valid_files = []
     for file in files:
         try:
-            hdul = fits.open(name=file, memmap=False, cache=False, lazy_load_hdus=False, ignore_missing_end=True)
-            valid_files.append(file)
+            with fits.open(name=file, memmap=False, cache=False, lazy_load_hdus=False, ignore_missing_end=True) as hdui:
+                valid_files.append(file)
         except OSError as e:
             log_info(f"Warning: corrupted file found and removed from reduction\n\t-File: {file}\n\t-Reason: {e}", warn=True)
-        finally:
-            if getattr(hdul, "close", None) and callable(hdul.close):
-                hdul.close()
-            del hdul
 
     return valid_files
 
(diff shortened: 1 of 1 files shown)


mikeprimm:
One additional thought here - the existing use of memmap=False is likely driving some very heavy native memory allocation demands: I'd expect the numpy array that is being created for the 'data' on the hdu is being forced to be a single large malloc(), versus being a mmap() of the data in the file, both forcing the whole array to be found as a contiguous array in the process heap, and forcing the whole content to be loaded into that array.  I see with some of the other fields that the goal is likely to force an integrity check by driving things to be loaded vs not, but I would wonder whether there is actually any integrity checking of the 'data' field itself beyond what would be involved in the mmap.
The driving scenario for this PR looked to be related to this - user with 100+ MB FITS files was blowing up about 50 files into running on Collab - file it popped on was fine, and worked elsewhere.  Still checking with the user, but if removing the otherwise good file just pushes the problem to file 51, this scenario seems likely.

mikeprimm:
I've made an addition to this, to work around a problem seen with very large data sets in Google Colab: specifically, the extreme intensity of the initial I/O of the whole set of FITS files done here, with a Google Drive as a source, seems to cause a huge but temporary virtual memory bubble in the Colab VM.  The additional code, besides addressing the initial problem, does a 5 second pause and retries once when an error is encountered (which might be due to a corrupt file OR due to a temporary network file system error due to exhausted virtual memory): if the file is actually corrupt, it still will be on the second try, but the 5 second delay seems sufficient to allow the bubble from the intense I/O to recover.  I'm running a few more tests with the known-problematic 40GB (371 x 120MB files), but the retry should, at worst, be non-impacting.

mikeprimm:
OK 4 for 4 clean runs on the full 40GB set - the retry jumped in to the rescue on 3 of them, so feeling pretty good.  Consistent good results :)

rzellem (approved):

PR #1170, opened 2023-04-23
Add logging of AAVSO comparison star additions

mikeprimm:
Very simple change - just to identify that each AAVSO contributed comparison star, both to indicate that it did happen and what pixel coordinate it corresponds to.

### DIFF
--- a/exotic/exotic.py
+++ b/exotic/exotic.py
@@ -836,6 +836,7 @@ def check_comps(comp_stars, vsp_comp_stars, tol=10):
 
         if not inlist:
             comp_stars_list.append(vsp_comp)
+            log_info(f"Added Comparison Star #{len(comp_stars_list)} [{vsp_comp[0]},{vsp_comp[1]}] from AAVSO")
 
     return comp_stars_list, vsp_pix
 


exotic/exotic.py
```diff
@@ -836,6 +836,7 @@ def check_comps(comp_stars, vsp_comp_stars, tol=10):
 
         if not inlist:
             comp_stars_list.append(vsp_comp)
+            log_info(f"Added Comparison Star #{len(comp_stars_list)} [{vsp_comp[0]},{vsp_comp[1]}] from AAVSO")
```
jpl-jengelke, line 839: I'm approving this but just wanted to check whether any of those values are tested to guarantee they exist before writing them in a log statement. This could potentially cause an unintended exception.

jpl-jengelke (approved):

exotic/exotic.py
mikeprimm: Relative to the line before, the existing code already makes this assumption (if vsp_comp were bad, it still winds up in the comp_star_list and would break later).  Specifically,  https://github.com/mikeprimm/EXOTIC/blob/07719b588c28b0e2759a528090f509a0e6b70d5b/exotic/exotic.py#L827 takes the provided list (which would break if it were undefined) and pulls out the 'xy' field, which is both what is printed but also what gets used downstream without further validation.

PR #1171, opened 2023-04-23
Add plate status logging, more friendly (less redundant) errors

mikeprimm:
This is intended to both act as a usability improvement (making the errors presented to the users for various frame related problems more oriented around what they need to know, versus a more programmer-centric error), and an improvement in the final processing artifacts for the run (specifically, the creation of a PlateStatus-<target>-<date>.csv file recording the various error/warning conditions for each science frame provided by the user - including those rejected due to FITS formatting, inability to be aligned versus the first plate, low flux, and out of frame conditions for the target and each comparison star).
Specifically, existing log messages were switched to log.debug (so they are there for programmer debug, but don't show in the user's console), while more friendly warning message are show (and deduplicated on each frame - having an out of frame error for every aperture size attempted on a given frame is excessive).  The status for all input files are ultimately recorded, and used to generate a final CSV file in the 'temp' directory, recording all the relevant frame-specific issues that may or may not have caused the frame to either be rejected or possibly not be fully utilized.

### DIFF
--- a/exotic/exotic.py
+++ b/exotic/exotic.py
@@ -124,6 +124,10 @@ try:  # output files
     from output_files import OutputFiles, VSPOutputFiles
 except ImportError:  # package import
     from .output_files import OutputFiles, VSPOutputFiles
+try:
+    from plate_status import PlateStatus
+except ImportError:
+    from .plate_status import PlateStatus
 try:  # plots
     from plots import plot_fov, plot_centroids, plot_obs_stats, plot_final_lightcurve, plot_flux, \
         plot_stellar_variability, plot_variable_residuals
@@ -151,7 +155,6 @@ plt.style.use(astropy_mpl_style)
 # logging -- https://docs.python.org/3/library/logging.html
 log = logging.getLogger(__name__)
 
-
 def log_info(string, warn=False, error=False):
     if error:
         print(f"\033[31m {string}\033[0m")
@@ -162,6 +165,8 @@ def log_info(string, warn=False, error=False):
     log.debug(string)
     return True
 
+# Initialze plate status log
+plateStatus = PlateStatus(log_info)
 
 def sigma_clip(ogdata, sigma=3, dt=21, po=2):
     nanmask = np.isnan(ogdata)
@@ -569,6 +574,7 @@ def nonlinear_ld(ld, filter_):
 def corruption_check(files):
     valid_files = []
     for file in files:
+        plateStatus.setCurrentFilename(file)
         try:
             with fits.open(name=file, memmap=False, cache=False, lazy_load_hdus=False, ignore_missing_end=True) as hdu1:
                 valid_files.append(file)
@@ -582,8 +588,8 @@ def corruption_check(files):
                 with fits.open(name=file, memmap=False, cache=False, lazy_load_hdus=False, ignore_missing_end=True) as hdu1:
                     valid_files.append(file)
             except OSError as e:
-                log_info(f"Warning: corrupted file found and removed from reduction\n\t-File: {file}\n\t-Reason: {e}", warn=True)
-
+                log.debug(f"Warning: corrupted file found and removed from reduction\n\t-File: {file}\n\t-Reason: {e}")
+                plateStatus.fitsFormatError(e)
     return valid_files
 
 def check_wcs(fits_file, save_directory, plate_opt, rt=False):
@@ -896,7 +902,8 @@ def transformation(image_data, file_name, roi=1):
                     except Exception:
                         pass
 
-    log_info(f"Warning: Following image failed to align - {file_name}", warn=True)
+    log.debug(f"Warning: Following image failed to align - {file_name}")
+    plateStatus.alignmentError()
     return SimilarityTransform(scale=1, rotation=0, translation=[0, 0])
 
 
@@ -1021,7 +1028,7 @@ def mesh_box(pos, box, maxx=0, maxy=0):
 
 
 # Method fits a 2D gaussian function that matches the star_psf to the star image and returns its pixel coordinates
-def fit_centroid(data, pos, psf_function=gaussian_psf, box=15, weightedcenter=True):
+def fit_centroid(data, pos, starIndex, psf_function=gaussian_psf, box=15, weightedcenter=True):
     # get sub field in image
     xv, yv = mesh_box(pos, box, maxx=data.shape[1], maxy=data.shape[0])
     subarray = data[yv, xv]
@@ -1029,7 +1036,8 @@ def fit_centroid(data, pos, psf_function=gaussian_psf, box=15, weightedcenter=Tr
         init = [np.nanmax(subarray) - np.nanmin(subarray), 1, 1, 0, np.nanmin(subarray)]
     except ValueError as ve:
         # Handle null subfield - cannot solve
-        log_info(f"Warning: empty subfield for fit_centroid at {np.round(pos, 2)}", warn=True)
+        plateStatus.outOfFrameWarning(starIndex)
+        log.debug(f"Warning: empty subfield for fit_centroid at {np.round(pos, 2)}")
         return np.empty(7) * np.nan
     # compute flux weighted centroid in x and y
     wx = np.sum(xv[0]*subarray.sum(0))/subarray.sum(0).sum()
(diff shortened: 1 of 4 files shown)


mikeprimm:
Here's a sample of the PlateStatus.csv:

`# filename,time,alignment_error,fits_error,lowflux_comp1,lowflux_comp2,lowflux_comp3,lowflux_target,outofframe_comp1,outofframe_comp2,outofframe_comp3,outofframe_target,skybg_comp1,skybg_comp2,skybg_comp3,skybg_target
"/home/mike/git/EXOTIC_sampledata/HatP32Dec202017/HATP-32171220013343.FITS",2458107.5654319096,False,False,False,False,False,False,False,False,False,False,False,False,False,False
"/home/mike/git/EXOTIC_sampledata/HatP32Dec202017/HATP-32171220013612.FITS",2458107.567157454,False,False,False,False,False,False,False,False,False,False,False,False,False,False
"/home/mike/git/EXOTIC_sampledata/HatP32Dec202017/HATP-32171220013912.FITS",2458107.569241412,False,False,False,False,False,False,False,False,False,False,False,False,False,False
"/home/mike/git/EXOTIC_sampledata/HatP32Dec202017/HATP-32171220014212.FITS",2458107.5713237035,False,False,False,False,False,False,False,False,False,False,False,False,False,False
"/home/mike/git/EXOTIC_sampledata/HatP32Dec202017/HATP-32171220014512.FITS",2458107.5734055787,False,False,False,False,False,False,False,False,False,False,False,False,False,False
"/home/mike/git/EXOTIC_sampledata/HatP32Dec202017/HATP-32171220014812.FITS",2458107.5754887382,False,False,False,False,False,False,False,False,False,False,False,False,False,False
`

exotic/output_files.py
```diff
@@ -130,7 +134,9 @@ def aavso(self, comp_star, airmasses, ld0, ld1, ld2, ld3, epw_md5):
                 f.write(f"{round(self.fit.time[aavsoC], 8)},{round(self.fit.data[aavsoC], 7)},"
                         f"{round(self.fit.dataerr[aavsoC], 7)},{round(airmasses[aavsoC], 7)},"
                         f"{round(self.fit.airmass_model[aavsoC], 7)}\n")
-
+    def plate_status(self, plate_status: PlateStatus):
+        plate_status_file = self.dir / "temp" / f"PlateStatus_{self.p_dict['pName']}_{self.i_dict['date']}.csv"
+        plate_status.writePlateStatus(plate_status_file)
```
jpl-jengelke, line 139: It would be good to have a try-except here. Lots of things can happen during a write, like insufficient disk permissions for the user or an i/o failure.

exotic/plate_status.py
```diff
@@ -0,0 +1,99 @@
...
+            self.statusByFilename[self.filename] = {}
+        self.statusByFilename[self.filename]['time'] = time
+        return self
+    # Reort frame time
+    def alignmentError(self):
+        self._logError("alignment_error",
+            f"File {self.filename} failed to align with first file")
+    # Write plate status to CSV file
+    def writePlateStatus(self, file: str):
+        with open(file, 'w') as f:
```
jpl-jengelke, line 87: I'm trying to use Pathlib for most things nowadays, but okay.

exotic/exotic.py
```diff
@@ -2014,7 +2035,7 @@ def main():
                 for a, aper in enumerate(apers):
                     for an, annulus in enumerate(annuli):
                         if not np.isnan(psf_data['target'][i, 0]):
-                            aper_data["target"][i][a][an], aper_data["target_bg"][i][a][an] = aperPhot(imageData,
+                            aper_data["target"][i][a][an], aper_data["target_bg"][i][a][an] = aperPhot(imageData, 0,
```
jpl-jengelke, line 2038: This is an API change for `aperPhot(...)` (note new parm). Are we sure that this was wrong or did the signature of the method also change? If the signature changed, can we be sure nothing else uses it?

exotic/exotic.py
```diff
@@ -1990,16 +2014,13 @@ def main():
...
 
                     for j, coord in enumerate(compStarList):
                         ckey = f"comp{j + 1}"
 
                         cx, cy = tform(coord)[0]
-                        if cx > 0 and cy > 0:
-                            psf_data[ckey][i] = fit_centroid(imageData, [cx, cy])
-                        else:
-                            psf_data[ckey][i] = np.empty(7) * np.nan
+                        psf_data[ckey][i] = fit_centroid(imageData, [cx, cy], j+1)
```
jpl-jengelke, line 2023: Does `fit_centroid(...)` return an `np.nan` by default when either of the 2-D array vals are negative? Is this something that can be safely written out of the code?

exotic/exotic.py
```diff
@@ -1957,7 +1984,7 @@ def main():
                         pix_coords = wcs_hdr.world_to_pixel_values(tar_radec[0], tar_radec[1])
                         tx, ty = pix_coords[0].take(0), pix_coords[1].take(0)
 
-                    psf_data['target'][i] = fit_centroid(imageData, [tx, ty])
+                    psf_data['target'][i] = fit_centroid(imageData, [tx, ty], 0)
```
jpl-jengelke, line 1987: Safe change to this call?

jpl-jengelke (commented):
Please address the questions/comments or report why they are invalid. ...

exotic/exotic.py
mikeprimm-mirific: well, the negative pixel coordinate check was ultimately a half solution, since you also get exceptions when the cx or cy exceed the positive dimensions, just the same (I've experienced these frequently).   The new logic handles both in the fit_centroid versus presuming that the caller can/will know all the conditions to be avoided.

exotic/exotic.py
mikeprimm-mirific: Not sure what is meant - it's consistently changed at all points in the code.

exotic/output_files.py
mikeprimm-mirific: Would not presume to disagree, but it IS consistent with every other output file, in that none of them to do try-catch in the output method, and instead wrap them in a try-catch in the calling code in exotic.py - e.g. https://github.com/rzellem/EXOTIC/blob/bfaf8ddacf86cff3adb326126e219cb3e9ab2f05/exotic/exotic.py#L2533 vs https://github.com/rzellem/EXOTIC/blob/bfaf8ddacf86cff3adb326126e219cb3e9ab2f05/exotic/exotic.py#L2525 vs https://github.com/rzellem/EXOTIC/blob/bfaf8ddacf86cff3adb326126e219cb3e9ab2f05/exotic/exotic.py#L2519 and others.    Honestly, like the Pathlib comment below - as a PR, I was not intending to second guess the existing coding styles and conventions, whether or not I specifically agree with them - my goal with a PR is minimum impact versus being 'style police' :)

exotic/plate_status.py
mikeprimm-mirific: See former comment - net-net, this is consistent with other file output code patterns versus what I personally might prefer, and the PR attempts to follow pattern versus prescribe changes that would only be inconsistent with existing code unless the change altered all of them, as well.

exotic/exotic.py
mikeprimm-mirific: Not clear on why this is singled out as an API versus not - this is an application, not a library, so if we consider certain functions to be published/supported APIs versus not, I certainly do not see any convention by which this is able to be determined.   If we need to preserve this binding, wrapping it is an option

exotic/output_files.py
jpl-jengelke: OK ...

exotic/exotic.py
jpl-jengelke: So the method signature was changed and every call to it was changed to correspond? Just checking.

exotic/exotic.py
jpl-jengelke: :heavy_check_mark:

exotic/plate_status.py
jpl-jengelke: ✔️

exotic/exotic.py
jpl-jengelke: Parts of it actually are imported (as a library) into internal applications, and potentially could be used that way outside, also. However, I don't think this part of the code is used that way by the Lab. Failsafe is that if anything breaks it can be pinned to the last version. Anyway, I was mostly considering any calls within the application to the same code. Also, the comment breaks the line so things look a little weird. Anyway, just a sanity check.

exotic/exotic.py
jpl-jengelke: @mikeprimm @mikeprimm-mirific So the method signature was changed and every call to it was changed to correspond? Just another sanity check. (See earlier reply about library usage.)

mikeprimm: Yep - scanned through and changed all references, so should be consistent throughout, and I've run tests on all relevant paths.

pearsonkyle (approved):

PR #1172, opened 2023-04-23
Handle times, jd_times arrays properly when first image files skipped

mikeprimm:
This is a simple fix for what is probably a rare scenario, but it should avoid a serious problem when it does occur: when the fix_centroid check on the first frames throws an error or otherwise does not match, and the bad frames are skipped, the pruning of the inputfiles array is done, but the already existing times and jd_times array does not have the corresponding pruning done - this would cause an off by N problem for all downstream processing, as well as the arrays not being of matching lengths during model processing and the like.
I've also added a log message for when this pruning does occur

### DIFF
--- a/exotic/exotic.py
+++ b/exotic/exotic.py
@@ -1853,6 +1853,7 @@ def main():
                     del first_image
 
             if inc > 0:
+                log_info(f"Skipping first {inc} files - Target star not found")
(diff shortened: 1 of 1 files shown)


jpl-jengelke (approved):
Approved. Looks solid. Please also comment @pearsonkyle @rzellem

PR #1173, opened 2023-04-26
Handle tFlux, cFlux sigma clipping during model matches

mikeprimm:
I believe this is behind some of the errors we've seen, relative to funky plots (with multiple lines) as well as exceptions on plot_flux calls (due to inconsistent array sizes/layouts).

Specifically, this is driven by fit_lightcurve, which can prune the provided times and flux values based on sigma clipping:

https://github.com/rzellem/EXOTIC/blob/develop/exotic/exotic.py#L1449

When this happens, the times and flux arrays used for the light curve can be reduced below the input population - these reduced versions are returned in myfit and as cFlux1 and tFlux1, respectively.  When no points are sigma clipped, these are the same as those input to the call (which is why things work most of the time).

The use of the UNPRUNED values later (such as https://github.com/rzellem/EXOTIC/blob/develop/exotic/exotic.py#L2148):

                        goodTargets = tFlux
                        goodReferences = cFlux
                        goodTUnc = tFlux ** 0.5
                        goodRUnc = cFlux ** 0.5
                        
works when no pruning has occurred, but breaks the assumption of length and index value consistency leveraged later (specifically, around further reordering and pruning via [si][gi] - generated here (https://github.com/rzellem/EXOTIC/blob/develop/exotic/exotic.py#L2271).  The use of these for plot_flux (https://github.com/rzellem/EXOTIC/blob/develop/exotic/exotic.py#L2322) will then result in index and length inconsistencies between times and the flux arrays.

The need for this is also supported by the existing use of cFlux1/tFlux1, which demonstrate the resulting index/length correlation.


mikeprimm:
Just noticed this is against main vs develop - closing and reopening PR against that branch

mikeprimm:
Reopened as https://github.com/rzellem/EXOTIC/pull/1183

PR #1175, opened 2023-05-01
Make times BJD-TDB immediately, remove later conversion

mikeprimm:
I'm interested in feedback here, but let me outline the theory and motive:  the existing logic for producing the 'times' versus 'jd_times' arrays for the frame times currently results in jd_times always being a Julian Date, while 'times' will be either a BJD-TDB OR a JD time (depending upon whether the FITS files happen to have or not have BJD headers).  Besides the concern of having those values be essentially of inconsistent definition, the result of this is that the times used in the model fits are (potentially) inconsistent with the time reference for the priors - specifically, the fit_lightcurve (https://github.com/rzellem/EXOTIC/blob/develop/exotic/exotic.py#L1442) is called before the existing code ultimately 'normalizes' the times to BJD-TDB (https://github.com/rzellem/EXOTIC/blob/develop/exotic/exotic.py#L2249), which results in the processing of the times there (relative to computing phase and such from 'tmid', which IS always BJD-TDB) being inconsistent when that processing is being done.  I'm not sure how much effect this will have on the light curve fit (it might be more of an issue when tmid has very tight tolerances, for example), but it would seem to cause any inappropriate (and unnecessary) bias on the match relative to the priors.

In any case, the PR attempts to make 'times' be strongly defined as BJD-TDB as soon as the FITS are loaded, removing the need to 'fix' the values later (and ultimately resulting in the essentially the same processing costs, in either case), while assuring that all model fits are being done 'apples to apples' on times versus priors.

### DIFF
--- a/exotic/exotic.py
+++ b/exotic/exotic.py
@@ -227,7 +227,7 @@ def get_exp_time(hdr):
     exp_time = next((exptime for exptime in exp_list if exptime in hdr), None)
     return hdr[exp_time] if exp_time is not None else 0.0
 
-def img_time(hdr, var=False):
+def img_time_jd(hdr):
     """Converts time from the header file to the Julian Date (JD, if needed)
     and adds an exposure offset (if needed)
 
@@ -235,9 +235,6 @@ def img_time(hdr, var=False):
     ----------
     hdr : astropy.io.fits.header.Header
         A header file that includes the time from when the image was taken
-    var : bool
-        Flag for if only JD is needed due to Stellar Variability code (ignore BJD)
-
     Returns
     -------
     float
@@ -249,7 +246,6 @@ def img_time(hdr, var=False):
         time_list = ['BJD_TDB', 'BJD_TBD', 'BJD'] + time_list
 
     exp = get_exp_time(hdr);
-
     hdr_time = next((time for time in time_list if time in hdr), None)
 
     if hdr_time == 'MJD_OBS':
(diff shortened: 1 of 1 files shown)


jpl-jengelke:
In general, my preference is to keep a generic function that defaults to BJD-TDB unless it's not detected in the headers, in which case it would default to JD, or alternately do the conversion immediately. Because I don't know the original intent of this code, I defer to @pearsonkyle for more analysis.

.gitignore
```diff
@@ -131,3 +131,4 @@ newfits.fits
 eaConf.json
 exotic.log.*
 pl_names.json
+/.project
```
jpl-jengelke, line 134: What is the `.project` used by?

mikeprimm: Eclipse work file - never part of source, and adding it to the .gitignore keeps it from being checked in

pearsonkyle:
The logic makes sense to me. I think the BJD conversion was brought into the project after the light curve fitting part and so it happened to come after. I agree we should normalize the times prior to fitting the light curve otherwise we'd have to run another conversion on the mid-transit time to ensure it has the proper unit.

pearsonkyle (approved):

rzellem:
@mikeprimm - can you please check and resolve this potential conflict in `exotic/exotic.py`? after that, I think we can *finally* merge! yay!!!

mikeprimm:
Yep - some overlap, so I'll get it resolved now and do a quick retest.

rzellem (approved):

mikeprimm:
OK - test run looks good - thanks for the merge!

PR #1180, opened 2023-05-25
Add handling of EXP header, and avoid AAVSO error for large fields

mikeprimm:
This is handling two conditions exposed by Andrew Serine's data (ref - https://uol-ets.slack.com/archives/CJJRP4QKV/p1685032317200029)

EXOTIC does not handle the EXP header as a possible exposure time source
For large fields of view, the AAVSO VSP API pops, due to the API not accepting requests for magnitudes about 12 if the FOV is beyond 180 arcmin
This change consolidates the exposure time parsing into a common routine, and shifts both places where these data are used to call this common parser. It adds EXP as the 3rd preferred header, and returns a 0.0 exposure time in the event of no valid headers (not sure about this last part - it should be 'good enough' for frame time, in general, but not sure otherwise - particularly relative to AAVSO output).

### DIFF
--- a/exotic/exotic.py
+++ b/exotic/exotic.py
@@ -217,6 +217,10 @@ def julian_date(hdr, time_unit, exp):
 
     return julian_time + offset
 
+def get_exp_time(hdr):
+    exp_list = [ "EXPTIME", "EXPOSURE", "EXP" ]
+    exp_time = next((exptime for exptime in exp_list if exptime in hdr), None)
+    return hdr[exp_time] if exp_time is not None else 0.0
 
 def img_time(hdr, var=False):
     """Converts time from the header file to the Julian Date (JD, if needed)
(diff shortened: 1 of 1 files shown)


jpl-jengelke:
Generally, I agree with this PR but need another set of eyes closer to that particular section of code.

exotic/exotic.py
```diff
@@ -217,6 +217,10 @@ def julian_date(hdr, time_unit, exp):
 
     return julian_time + offset
 
+def get_exp_time(hdr):
+    exp_list = [ "EXPTIME", "EXPOSURE", "EXP" ]
+    exp_time = next((exptime for exptime in exp_list if exptime in hdr), None)
+    return hdr[exp_time] if exp_time is not None else 0.0
```
tamimfatahi, line 223: Looks good to me. I'm not too sure if we should leave the default case to 0 seconds, but I can't think of a better alternative currently. Nice find!

rzellem (approved):

PR #1189, opened 2023-06-14
Updates for Multi_Fit notebook + Periodogram

pearsonkyle:


examples/tess.py
```diff
@@ -739,27 +739,8 @@ def check_std(time, flux, dt=0.5): # dt = [hr]
 
         # create AAVSO formatted csv
         csv_lk = OutputFiles(myfit, prior, infoDict, planetdir)
-        try:
-            csv_lk.aavso_csv(airmass,u0,u1,u2,u3,tmidstr)
-            csv_lk.aavso(airmass,u0,u1,u2,u3, tmidstr)
-        except:
-            print(f"Failed to create AAVSO csv for {args.target} - Sector {lcdata['sector']}")
+        csv_lk.aavso_csv(airmass,u0,u1,u2,u3,tmidstr)
+        csv_lk.aavso(airmass,u0,u1,u2,u3, tmidstr)
```
rzellem, line 743: any reason to remove this try/except, @pearsonkyle ?

exotic/api/nested_linear_fitter.py
```diff
@@ -333,11 +335,12 @@ def plot_periodogram(self):
 
         ########################################
         # subtract first order solution from data and recompute periodogram
-        maxper = 50 # constrain second order solution to be less than 50 days
+        maxper = maxper2 # constrain second order solution to be less than 50 days
```
rzellem, line 338: should this comment be updated?

exotic/api/output_aavso.py
```diff
@@ -112,7 +112,6 @@ def aavso(self, airmasses, ld0, ld1, ld2, ld3, tmidstr):
                     f"#PRIORS-XC={dumps(priors_dict)}\n"  # code yields
                     f"#RESULTS=Tc={round_to_2(self.fit.parameters['tmid'], self.fit.errors['tmid'])} +/- {round_to_2(self.fit.errors['tmid'])}"
                     f",Rp/R*={round_to_2(self.fit.parameters['rprs'], self.fit.errors['rprs'])} +/- {round_to_2(self.fit.errors['rprs'])}"
-                    f",a/R*={round_to_2(self.fit.parameters['ars'], self.fit.errors['ars'])} +/- {round_to_2(self.fit.errors['ars'])}"
```
rzellem, line 115: why did you remove a/Rs here? because we don't fit for it explicitly?

rzellem (changes requested):

examples/tess.py
pearsonkyle: It is not necessary and complicates debugging if you can't get a traceback on the error. The errors have been fixed and the code runs without it

exotic/api/output_aavso.py
pearsonkyle: Yes, a/Rs is not always a free parameter and this line of code will error out. I've switched to fitting for inclination in order to estimate the transit duration

pearsonkyle: this line was causing that try/except statement above to trigger the except due to my comment above. It's not necessary to include it since we already list the value in the priors.

exotic/api/nested_linear_fitter.py
pearsonkyle: removed the comment... the period grid should likely be hand tailored anyway based on need.. it's tough to choose a one size fits all range

rzellem (approved):

PR #1203, opened 2023-07-21
Upgrade EXOTIC to Cython 3 -- testing required

### LINKED ISSUE #1202: Upgrade to Cython 3
[Changes are fairly extensive in the Cython 3.x upgrade (version 0.x to 3.x)](https://cython.readthedocs.io/en/latest/src/changes.html) so any detailed testing would be beneficial to validate with the updated library.

This should be done for obvious reasons but also because Cython 3 improves compatibility with newer Python versions. 

Currently some of EXOTIC's dependencies are relying on Cython 3, so it would be beneficial to unify the dependency tree by upgrading EXOTIC to avoid any unforeseen, weird interactions between old and new libraries.  The place to update Cython is `pyproject.toml` and the change (sans testing) is in this ticket's eponymously named branch.

jpl-jengelke:
This is a major upgrade of Cython from 0.x to 3.x -- It will require detailed testing to make sure nothing breaks. Since Ultranest moved to Cython 3, EXOTIC should move to it, also. Note that the version of Cython was left open-ended in `pyproject.toml` metadata (`requires`) to always imply the latest, which is the 3.x series. This can be modified later to pin Cython, as needed, but should be open-ended for now since the product is new and anticipates fixes. 

Closes #1202


tamimfatahi:
EXOTIC only uses Cython due to Ultranest. Since Ultranest already released an update regarding the Cython 3.0 upgrade (v3.6.2) and we made that our requirement, we should be set as we tested on multiple reductions.

tamimfatahi (approved):

PR #1211, opened 2023-08-05
Filter logic modifications to extend matching

### LINKED ISSUE #1210: Extend filter matching for malformed values

jpl-jengelke:
Closes #1210 

- Remove redundancy in naming by auto-generating some value data
- Modify LimbDarkening OOP design to utilize static methods and create appropriate class variables
- Create matcher map to match against punctuation free and spacing normalized user input variables
- Extend filter matching to algorithmically examine filter inputs step-wise from most- to least-critical
   - Add `name` option to input filter
- Modify instance setters to set instance filters from class variables

### DIFF
--- a/exotic/api/filters.py
+++ b/exotic/api/filters.py
@@ -1,79 +1,78 @@
 # Sources for FWHM band wavelengths are referenced below, respectively 
 # note that all below units default to nm
-fwhm = {
+__fwhm = {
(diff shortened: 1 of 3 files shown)


jpl-jengelke:
Tested on my local workstation and passed. ...

rzellem (approved):

PR #1212, opened 2023-08-07
Follow on to Filter logic modifications to extend matching

### LINKED ISSUE #1210: Extend filter matching for malformed values

jpl-jengelke:
- Add method to validate wavelength values (FWHM) for optical filters
- Documentation
- Rev version in advance of release

Close #1210


jpl-jengelke (commented):
These appear as designed.

rzellem (approved):

PR #1214, opened 2023-08-08
Minor bug fixes and unity in filter detection

jpl-jengelke:
- Limit loose filter description matching
- Bug fix involving name detection
- Break out reserved filter abbreviations into class-level variables
- Other minor improvements and bug fixes


tamimfatahi (approved):
Looks good!

PR #1215, opened 2023-08-08
Minor bug fixes and unity in filter detection

### LINKED ISSUE #1210: Extend filter matching for malformed values

jpl-jengelke:
- Limit loose filter description matching
- Bug fix involving name detection
- Break out reserved filter abbreviations into class-level variables
- Other minor improvements and bug fixes

Closes #1210


tamimfatahi (approved):
Looks good!

PR #1217, opened 2023-08-08
Issue 1165: Checking datasets for MObs and Conforming w/ New Limb Darkening Updates

### LINKED ISSUE #1165: Default filter to MObs CV for MObs data
In the header, check if the data came from MObs. If so, default the filter to MObs CV. Might need to rearrange when we check the header for MObs as it occurs after the LD coefficients are calculated.

tamimfatahi:
- Added testing suite for `ld.py`
- Modified EXOTIC to handle updated changes of ld.py for standard filters
- If user is using MicroObservatory datasets, `MObs CV` filter will be used as default
- Organized code

Two things I'd like to fix before making this a PR are the following:
1. I'm currently working on getting standard filters with differing FWHM (ex: `CV`) to work properly. I should have this done soonish.
2. I don't agree with the logic of the following in[ ld.py](https://github.com/rzellem/EXOTIC/blob/14c3d6fc17b5cada406c93a4e684bec75376f152/exotic/api/filters.py#L75):
`"Clear (unfiltered) reduced to V sequence": "MObs CV"`
It doesn't make sense since not all clear filters are MicroObservatory ones. Maybe we remove this? Although, the description would be nice to keep to describe a `non-MObs CV` filter on the results page if we could keep it.

Closes #1165


exotic/api/ld.py
```diff
@@ -207,119 +207,13 @@ def set_ld(self, ld0, ld1, ld2, ld3):
...
-    test_ld(ld_obj, filter_info1)
-    # test_ld(ld_obj, filter_info2)
-    # test_ld(ld_obj, filter_info3)
-    # test_ld(ld_obj, filter_info4)
-
-    ld = [ld_obj.ld0[0], ld_obj.ld1[0], ld_obj.ld2[0], ld_obj.ld3[0]]
-    ld_unc = [ld_obj.ld0[1], ld_obj.ld1[1], ld_obj.ld2[1], ld_obj.ld3[1]]
-    ld_obj.output_ld()
+    def __str__(self):
+        return (f"\nFilter Name: {self.filter_name}"
```
jpl-jengelke, line 211: In this case, the 'Filter Name' is actually the 'desc' and what's called name is actually 'Filter Abbreviation'... The values should be reversed.

jpl-jengelke (changes requested):

exotic/api/ld.py
jpl-jengelke: I just made the change, and also removed the anachronistic `join` calls.

jpl-jengelke (approved):

exotic/api/tests/test_ld.py
jpl-jengelke, line 1: To prevent possibly packaging tests in the wheel, I think these should be in the `tests` directory at the top level. It will also encapsulate them away from the actual application code.

jpl-jengelke (changes requested):
Please move tests to the `tests` directory at the top level of the application. We can explore PyUnit to run them as an add-on later.

jpl-jengelke:
@rzellem Please comment/approve or add insight to the following ...
> I don't agree with the logic of the following in[ ld.py](https://github.com/rzellem/EXOTIC/blob/14c3d6fc17b5cada406c93a4e684bec75376f152/exotic/api/filters.py#L75):
"Clear (unfiltered) reduced to V sequence": "MObs CV"
It doesn't make sense since not all clear filters are MicroObservatory ones. Maybe we remove this? Although, the description would be nice to keep to describe a non-MObs CV filter on the results page if we could keep it.

As background, that is an alias to the MObs CV filter. Obviously we are using MObs CV as a default for clear in most cases. I think maybe @tamimfatahi is objecting to the generic wording being applied to a specific type of clear filter.

As an aside, if `CV` is entered and the `fwhm` vals don't match up to those for `MObs CV` do we still call it `Custom` with an abbreviation code of `N/A`? I believe AAVSO requested that all custom filters be tagged with 'N/A'. 

@tamimfatahi may have more prescient concerns. Thanks.

exotic/api/tests/test_ld.py
tamimfatahi: Done.

tamimfatahi:
With these changes, the `MObs CV` filter will no longer be linked to `CV`. The good news is that users will **not** have to change anything as we will detect if their images came from MicroObservatory, so it won't matter what they list as their filter (as long as we can tell based on their header).

exotic/api/filters.py
```diff
@@ -43,6 +43,9 @@
         "PanSTARRS Y": {"name": "Y", "fwhm": ("946.4", "1054.4")},
         "PanSTARRS w": {"name": "N/A", "fwhm": ("404.2", "845.8")},
 
+        "Clear (unfiltered) reduced to V sequence": {"name": "CV", "fwhm": (None, None)},
+        "Clear with blue-blocking": {"name": "CBB", "fwhm": (None, None)},
+
```
jpl-jengelke, line 48: These would be the only values in the known list without definitive FWHM values. That sort of makes it hard to use as a definitive reference. 

If you would like to use the names, I think we should either enter the user-set description and keep their entered filter values. I think the matching can spiral as there all kinds of paths that follow from heuristics. Heuristics is really a road to perdition. 

Another concern I have about CBB is leaving the fwhm values open ended. I believe `blue-blocking` implies specific wavelengths, so if we leave that open ended, someone could enter red blocking values or anything to make the text meaningless. 

On the CITISENS side I am dealing with all kinds of inputs from various platforms, including AstroImageJ, which does not do this kind of validation AFAIK. But there could still be the same thing done manually in EXOTIC. ... I guess I need to chat to understand the impetus for what is being attempted.

exotic/api/filters.py
```diff
@@ -72,7 +75,5 @@
         "LCO Pan-STARRS Y": "PanSTARRS Y",
         "LCO Pan-STARRS w": "PanSTARRS w",
 
-        "Clear (unfiltered) reduced to V sequence": "MObs CV",
```
jpl-jengelke, line 75: I support this if we are allowing a more liberal custom filter arrangement. 

There's no cost to matching the name, though. If we want to maintain a different 'name' (e.g. abbreviation) I think we should approach that with a slightly different algorithm. ... We can talk about it tonight if you're free.

jpl-jengelke:
Just as an aside, this is important for the TESS processing because precision is more critical given the nature of space-based telescope observations.

jpl-jengelke:
I have some ideas and was considering pseudocode right here, but I think we would be more proficient doing it together again tonight. Generally, if we want to preserve user descriptions for certain filter 'name' (e.g. abbreviation) values, I think we should consider that after all the matching, and on very limited bases for the reason mentioned above, wavelengths should generally match filter descriptions.

jpl-jengelke (changes requested):

exotic/api/ld.py
```diff
@@ -118,9 +116,15 @@ def check_standard(self, filter_: dict = None, loose: bool = False, loose_len: i
...
+                    if filter_['name'] in LimbDarkening.filter_nonspecific_ids or \
+                            (filter_['filter'] and filter_['filter'][:5].upper() in LimbDarkening.filter_nonspecific_ids):
+                        if filter_['wl_min'] and filter_['wl_max'] and LimbDarkening.check_fwhm(filter_):
+                            filter_alias = f
+                            filter_alias['fwhm'] = (filter_['wl_min'], filter_['wl_max'])
+                            break
                     # exclude unknown vals for 'name'
-                    if filter_['name'] in filter_nonspecific_names or \
-                            (filter_['filter'] and filter_['filter'][:5].upper() in filter_nonspecific_names):
+                    elif filter_['name'] in LimbDarkening.filter_generic_ids or \
```
jpl-jengelke, line 126: Recommend reverting this stuff. The reason being is that AAVSO recommends ignoring the full list of `nonspecific` names. So that means it would only match MObs CV is either the desc said `MObs CV` or the FWHM vals match. So I would perform a check against those `nonspecific` values at the conclusion of the testing and then fill in the blanks for the description as necessary. ... We could even make a map to replace the description if it's empty.

tamimfatahi:
@jpl-jengelke Updated for the latest addition. EXOTIC will still need to be modified to make this work, but for now, this PR can be approved.

tamimfatahi:
@jpl-jengelke PR should be good to go for release. Tested it on multiple different runs and outputs as expected. EXOTIC  now works with  the new additions.

PR #1219, opened 2023-08-22
Additional filter modifications

jpl-jengelke:
Moved lists of generalized filters to the filter module. Update matching against not formally specified filters to use sets. Some additional cleanup. ...


jpl-jengelke:
@tamimfatahi Is this good to go? I mean, no other changes?

Issue #1067, opened 2022-10-15
Variable Star Error for Multiple Transits


tamimfatahi:
When feeding in multiple transits, the magnitude of the star isn't able to be calculated.

tamimfatahi:
Completed in PR #1242

PR #1242, opened 2023-12-01
Issue 1240, 1094, 1067: Bringing back AID Uploads

tamimfatahi:
Currently, EXOTIC is not able to upload variable star measurements to AAVSO. This PR is a draft and includes all changes from open PRs I have opened. Please do not accept this PR until I get the correct results and all other open PRs have been closed.


tamimfatahi:
Added final commit to properly calculate the target's magnitude. Also, closes issue #1067

tamimfatahi:
Some highlights/results:

I tested this out on Leia, which used an R filter when I observed WASP-1 b. I've attached the stellar variability plot and AID upload timeseries file.

WASP-1 seems to have a M_SR = 11.62 according to VSX (Sloan R was the closest red-band I could find) and EXOTIC measured it to be M_Rc= ~11.28 (Cousins R).

Stellar_Variability
[AID_AAVSO_WASP-1_16-September-2023.txt](https://github.com/rzellem/EXOTIC/files/13620315/AID_AAVSO_WASP-1_16-September-2023.txt)

exotic/plots.py
```diff
@@ -140,19 +140,19 @@ def plot_flux(times, targ, targ_unc, ref, ref_unc, norm_flux, norm_unc, airmass,
 def plot_variable_residuals(save):
     plt.title("Stellar Variability Residuals")
     plt.ylabel("Residuals (flux)")
-    plt.xlabel("Time (BJD_TDB)")
+    plt.xlabel("Time (JD)")
```
rzellem, line 143: Isn't this still in BJD_TDB? We do the conversions ourselves to BJD_TDB, so I think this label is still accurate.

exotic/plots.py
```diff
@@ -140,19 +140,19 @@ def plot_flux(times, targ, targ_unc, ref, ref_unc, norm_flux, norm_unc, airmass,
...
-def plot_stellar_variability(vsp_params, save, s_name, label):
+def plot_stellar_variability(vsp_params, save, s_name, vsp_auid_comp):
     for vsp_p in vsp_params:
         plt.errorbar(vsp_p['time'], vsp_p['mag'], yerr=vsp_p['mag_err'], color="tomato", fmt='.')
 
-    plt.title(f"{s_name} (Label: {label})")
+    plt.title(f"{s_name} (Label: {vsp_auid_comp})")
     plt.ylabel("Vmag")
-    plt.xlabel("Time (BJD_TDB)")
+    plt.xlabel("Time (JD)")
```
rzellem, line 155: Isn't this still in BJD_TDB? We do the conversions ourselves to BJD_TDB, so I think this label is still accurate.

rzellem (changes requested):

exotic/plots.py
tamimfatahi: AAVSO requires us to save the it in the JD time format. I saved the corresponding JD times to the BJD ones. For reference, here is [AAVSO's](https://www.aavso.org/aavso-extended-file-format) available time formats:

`
- **DATE**: The format of the date used in the report. Times are midpoint of the observation. Convert all times from 
- UT to one of the following formats:
- JD: Julian Day (Ex: 2454101.7563)
- HJD: Heliocentric Julian Day
- EXCEL: the format created by Excel's NOW() function (Ex: 12/31/2007 12:59:59 a.m )
`

exotic/plots.py
tamimfatahi: AAVSO requires us to save the it in the JD time format. I saved the corresponding JD times to the BJD ones. For reference, here is [AAVSO's](https://www.aavso.org/aavso-extended-file-format) available time formats:

> - **DATE**: The format of the date used in the report. Times are midpoint of the observation. Convert all times from UT to one of the following formats:
> 1. JD: Julian Day (Ex: 2454101.7563)
> 2. HJD: Heliocentric Julian Day
> 3. EXCEL: the format created by Excel's NOW() function (Ex: 12/31/2007 12:59:59 a.m )

exotic/plots.py
rzellem: Hm. Weird. I would think that BJD_TDB is still ok if it is asking for HD.

Can you please ping some folks on our Slack (or let me know if you'd prefer me to take care of this)? Looks like Chuck C and Daniel Z would know the answer to this question, and if not, Bert Pablo should.

tamimfatahi: Bert replied the following when asked if we can use BJD_TDB:
> it is not. We accept HJD and JD but we don’t have a way to accept anything else currently

rzellem (approved):

pearsonkyle (approved):

PR #1243, opened 2023-12-02
Issue 1019: Scraping V-mag from NEA

### LINKED ISSUE #1019: Add V-mag to NEA scraper

tamimfatahi:
I shelved this for a bit; I don't know if we still need it. I remember we were implementing it when we were coding the variable star portion. If so, we can accept this and move to the develop branch. Closes #1019


jpl-jengelke:
@tamimfatahi So this is not on the plate for EXOTIC 4.4.0?

PR #1244, opened 2024-02-02
Stellar variability, comp star scraping, plate solution updates and various EXOTIC fixes

rzellem:

PR #1253, opened 2024-04-17
Fix for AAVSO comp additions - particularly when one matches target star

mikeprimm:
This is a fix for what was ultimately two issues:
1) When comps are being validated and added to, the comp list exotic_infoDict['comp_stars'] is manipulated, adding and removing records.  A second variable, ra_dec_wcs, containing the RA/Dec of each comp is also defined (and needs to be index-matched) - this array was added to by success of check_for_variable_stars() (which either removed the bad comp from exotic_infoDict['comp_stars'], or added the WCS to ra_dec_wcs), but the vsp_query() call - which adds AAVSO comps, only added to exotic_infoDict['comp_stars'], resulting in an array length mismatch that could throw error at end of photometry.
2) As it turns out, there ARE AAVSO comp stars that have planets - specificially XO-2 N.  The validation of AAVSO returned comps does not consider this - resulting in quite amusing results when the target matches most excellently with itself!

In any case, the solution consists of:
1) Stop building the ra_dec_wcs before the comp list is finalized - it's not needed, and that keeps things easier, as far as assuring index consistency
2) Pass in the target to vsp_query, so that it can be considered, along with the user comps, when evaluating whether an AAVSO provided comp is to be used

### DIFF
--- a/exotic/exotic.py
+++ b/exotic/exotic.py
@@ -826,8 +826,14 @@ def vsx_variable(ra, dec, radius=0.01, maglimit=14):
     except Exception:
         return False
 
+def build_comp_ra_dec(ra_wcs, dec_wcs, comp_stars):
+    comp_ra_dec = []
+    for _, comp_star in enumerate(comp_stars[:]):
+        comp_ra_dec.append([ra_wcs[int(comp_star[1])][int(comp_star[0])],
+            dec_wcs[int(comp_star[1])][int(comp_star[0])]])
+    return comp_ra_dec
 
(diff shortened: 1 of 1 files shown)


tamimfatahi:
Apologies if this PR hasn't been properly reviewed. I'll review the changes you implemented once PR #1244 gets accepted, as there will be merge conflicts between the two.

mikeprimm:
Resynchronized and retested fix with current head for 'develop' branch

tamimfatahi:
Looks good

tamimfatahi (approved):
Looks good! Thanks @mikeprimm

PR #1257, opened 2024-06-07
Add tess_individ_lc.py for individual TESS time series extraction and fitting without global fit.

noguerfederico:
This pull request adds the tess_individ_lc.py script, which allows for the extraction and fitting of individual exoplanet transits from TESS data without performing a global fit in the process. 
This script improves efficiency for users who need quicker access to individual light curve data to perform a joint fit or individual light curve analysis with other Python tools. 
Works with confirmed exoplanets.  
This was created from tess.py.  In tess.py the global fit is performed first and the derived parameters are used as priors in the individual fits. In tess_individ_lc.py since we do not perform the global fit, the NEA priors are used.  This may give slightly different results on the individual fits.    
Outputs go to "tess/tess_individ_lc_output" folder.

### DIFF
--- a/examples/tess/tess_individ_lc.py
+++ b/examples/tess/tess_individ_lc.py
@@ -1,0 +1,697 @@
+
+# first create an environment with the proper dependencies + version
+# git clone https://github.com/rzellem/EXOTIC.git
+# cd EXOTIC
+#git status to check branch
+# git pull uodates changes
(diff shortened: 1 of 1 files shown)


noguerfederico:
Switching to working directly in the original repository.

jpl-jengelke:
Tracking this the actual merge occurred in PR #1262. No need to track further.

Issue #1264, opened 2024-07-23
PlateSolution class is timing out on users


IraLeeBell:
As of late, multiple users have experienced a timeout when the PlateSolution class is leveraged. Creating an issue here, so that I can tie this to a branch. Current thinking is to test increasing number of requests before giving up and potentially also specifying a timeout period explicitly. If this should work in my testing environment, I'll show a few people how to point to a test branch of the exotic repo to call any changes we've made in plate_solution.py, exotic.py, or elsewhere.

IraLeeBell:
The fix seems to be working. Several members of the community are testing now. Will report back further and make any adjustments prior to doing a pull request.

PR #1270, opened 2024-08-07
Plate solution timeout extension ...

jpl-jengelke:
This is a redux of #1267 that merges #1264 into `develop`. The redo is to ensure the correct release path for tracking purposes.


jpl-jengelke:
Minor change to be added to this PR to abstract the timeout into a variable in the Python class. Also, increasing Tenacity retries to 30 is a recipe for freezes since the timeout is set exponential (factor 2^t), anyway. TBC w/in next two days.

rzellem (approved):

Issue #1290, opened 2024-08-14
Colab notebooks fail on EXOTIC 4.0 due to breaking Bokeh API changes


jpl-jengelke:
Bokeh error in Colab using Jeff's latest version:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-1-f934c5145974> in <cell line: 89>()
     88 
     89 try:
---> 90   from exotic.api.colab import *
     91   from exotic.api.plotting import plot_image
     92 except ImportError:
/usr/local/lib/python3.10/dist-packages/exotic/api/colab.py in <module>
     58 # from bokeh.io import output_notebook
     59 from bokeh.palettes import Viridis256
---> 60 from bokeh.plotting import figure, output_file, show
     61 from bokeh.models import BoxZoomTool, ColorBar, FreehandDrawTool, HoverTool, LinearColorMapper, LogColorMapper, \
     62   LogTicker, PanTool, ResetTool, WheelZoomTool

/usr/local/lib/python3.10/dist-packages/bokeh/plotting/__init__.py in <module>
     20 # this is just for testing, otherwise the figure module is shadowed
     21 # by the figure function and inacessible (needs happen up top)
---> 22 from . import figure as _figure ; _figure
     23 
     24 #-----------------------------------------------------------------------------

/usr/local/lib/python3.10/dist-packages/bokeh/plotting/figure.py in <module>
     61 from ._stack import double_stack, single_stack
     62 from ._tools import process_active_tools, process_tools_arg
---> 63 from .glyph_api import _MARKER_SHORTCUTS, GlyphAPI
     64 
     65 #-----------------------------------------------------------------------------

/usr/local/lib/python3.10/dist-packages/bokeh/plotting/glyph_api.py in <module>
     46 #-----------------------------------------------------------------------------
     47 
---> 48 class GlyphAPI:
     49     """ """
     50 

/usr/local/lib/python3.10/dist-packages/bokeh/plotting/glyph_api.py in GlyphAPI()
    491 """
    492 
--> 493     @glyph_method(glyphs.Oval)
    494     def oval(self, *args: Any, **kwargs: Any) -> GlyphRenderer:
    495         """

AttributeError: module 'bokeh.models.glyphs' has no attribute 'Oval'
```
... and ... 
```
/content/exotic_env/lib/python3.10/site-packages/exotic/api/colab.py in display_image(filename)
    110 
    111     # create a figure with text on mouse hover
[...]

PR #1292, opened 2024-08-14
Issue 1290: Hotfix for Colab Bokeh breaking API issues

jpl-jengelke:
Minor API changes with a big impact.

- These will be tagged as a point release, and it is meant to accompany a newer notebook release. 

[EXOTIC Colab Notebooks (`exotic-colab`)](https://github.com/alienlifeform/exotic-colab) are the accompanying part of EXOTIC that runs on Google Colab. Anyone using an older Colab notebook (pre-EXOTIC 4.0.1 release), either off-the-shelf or customized, should default to [EXOTIC 3.3.0](https://github.com/rzellem/EXOTIC/releases/tag/3.3.0) either in this repository (above link) or at [exotic PyPi](https://pypi.org/project/exotic/3.3.0/).

The newer notebook benefits by unpinning the Bokeh image presentation module, see #1290 .


exotic/api/colab.py
```diff
@@ -109,7 +109,7 @@ def display_image(filename):
     # )
 
     # create a figure with text on mouse hover
-    fig = figure(tooltips=[("x", "$x"), ("y", "$y"), ("value", "@image")], plot_width=p_width, plot_height=p_height,
+    fig = figure(tooltips=[("x", "$x"), ("y", "$y"), ("value", "@image")], width=p_width, height=p_height,
```
rzellem, line 112: @jpl-jengelke - should this be a try/except instead to ensure backwards compatibility with the previous version of bokeh?

exotic/api/plotting.py
```diff
@@ -118,7 +118,7 @@ def plot_image(filename, save=False, bg_min=60, bg_max=99):
 
     # create a figure with text on mouse hover\
     print("Saturated pixels are marked with red. These are pixels which have exceeded the maximum value for brightness, and are thus not suitable for use as comparison stars.")
-    fig = figure(tooltips=[("x", "$x"), ("y", "$y"), ("value", "@image")], plot_width=800, plot_height=800,
+    fig = figure(tooltips=[("x", "$x"), ("y", "$y"), ("value", "@image")], width=800, height=800,
```
rzellem, line 121: @jpl-jengelke - should this be a try/except instead to ensure backwards compatibility with the previous version of bokeh?

rzellem (changes requested):

ivenzor:
I added try/except in my fork, currently testing.

exotic/api/colab.py
jpl-jengelke: Well, I thought no. Since it will be tagged as 4.0.1, we just need to make sure that it is only used to accompany a new Colab notebook. So releasing an upgraded notebook would be the solution.

exotic/api/plotting.py
jpl-jengelke: Well, I thought no. Since it will be tagged as 4.0.1, we just need to make sure that it is only used to accompany a new Colab notebook. So releasing an upgraded notebook would be the solution. 

In other words, it's isolated from the 3.3.0 release, so ideally shouldn't be a problem. Moreover, the modules need to be upgraded so supporting old modules wouldn't comport with security recommendations.

ivenzor:
If/when this is merged in main, the retail EXOTIC Standard Colab version must also be changed, these lines must be commented out:
```
#if bokeh.__version__ != "2.4.3":
#  !pip install bokeh==2.4.3
#  hideWarning()
#else:
#  showProgress(.3)
```

rzellem (approved):

PR #1293, opened 2024-08-14
Add try except for backward compatibility for older versions of bokeh

ivenzor:
Following Rob recommendation, try/excepts are added for backward compatibility


ivenzor:
@jpl-jengelke I guess this is no longer wanted, I will go ahead and close the pull request.

PR #1306, opened 2024-09-27
Update Trusted Publishing spec to fix bugs

jpl-jengelke:
- Fix path spec
- Add hash and attestation publishing


tamimfatahi (approved):
Looks good to me.

PR #1307, opened 2024-09-27
Issue 1288: Adding CodeQL and Ruff as additional workflow items

tamimfatahi:
CodeQL will help identify vulnerability issues, and Ruff as a linter to detect errors and formatting suggestions.

- I decided to replace Pylint with [Ruff](https://docs.astral.sh/ruff/) as many major projects are utilizing it due to speed.
- If you want the configuration file for CodeQL added rather than having it queried, let me know and I can add it in.


.github/workflows/codeql.yml
jpl-jengelke, line 1: I think it is missing the `codeql` directory, which is the configuration, from [here](https://github.com/NASA-AMMOS/slim-starterkit-python/tree/main/.github/workflows).

.github/workflows/ruff.yml
```diff
@@ -0,0 +1,69 @@
...
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [develop]
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: write
```
jpl-jengelke, line 27: Does it need write access? That seems like we're handing the VM a lot of power.

.github/workflows/ruff.yml
```diff
@@ -0,0 +1,69 @@
...
+    - name: Checkout repository
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.10'
+    - name: Upgrade tooling
+      run: |
```
jpl-jengelke, line 40: Is there a way to use the build command to make this happen? Maybe through a virtual environment so there's not so much installation stuff (that seems a little friable for future upgrades)? Please let me know if I'm wrong.

.github/workflows/ruff.yml
```diff
@@ -0,0 +1,69 @@
...
+        do
+            pythonpathplus="${pythonpathplus:+:${pythonpathplus}}:$p"
+        done
+        echo "PYTHONPATH=${PYTHONPATH:+:${PYTHONPATH}}${pythonpathplus}:." >> $GITHUB_ENV
+    - name: Analyze
+      # Refer to https://docs.astral.sh/ruff/configuration/#command-line-interface
+      # to add extra rules or a configuration file.
+      run: ruff check --output-file ruff_report.txt . || true
+      continue-on-error: true
+    - name: Upload Artifact
```
jpl-jengelke, line 62: Looks great. But what if we don't compress it? Then maybe it can be viewable in the browser instead of requiring a bunch of clicks and unpacking, etc.

.github/workflows/ruff.yml
```diff
@@ -0,0 +1,69 @@
...
+    - name: Upgrade tooling
+      run: |
+        python3 -m pip install --upgrade pip
+        pip3 install --upgrade build importlib_metadata setuptools setuptools_scm wheel
+        pip3 install ruff
+    - name: Install dependencies
+      run: |
+        pip3 install -r requirements.txt
+        pip3 install -e .
+    - name: Prepare PYTHONPATH
```
jpl-jengelke, line 48: All that path stuff is required, I think. Well maybe it can be condensed into the above: 'Prepare Test Environment' or something...

jpl-jengelke (changes requested):

.github/workflows/ruff.yml
tamimfatahi: It does not as it's merely analyzing the code. I'm assuming I should take away its permission to write during security-events as well?

.github/workflows/ruff.yml
tamimfatahi: We can run `ruff check --output-format=github .` and have it output in the browser if you'd like.

.github/workflows/ruff.yml
tamimfatahi: Make what happen exactly? Upgrading tools? I'm a little confused what installing tools through a virtual environment would do in this situation and how it would mitigate the need to install extra stuff. 

Also, I don't think we need `pip3 install --upgrade build importlib_metadata setuptools setuptools_scm wheel` as we're running a lint check on the code.

.github/workflows/codeql.yml
tamimfatahi: I'll make sure to add them in.

.github/workflows/python-publish.yml
jpl-jengelke, line 1: Let's talk about this one please. ... it's sort of a prototype for shared configurations, so needs things to be a little generic.

jpl-jengelke:
I will hold this for the next release. We have a lot of balls up in the air. Thanks!

opusbuilds:
Executed review, 2026-08-21 (part of the open-PR triage Rob asked me to run; evidence only, decisions stay with maintainers).

**What it does:** adds CodeQL security scanning and Ruff linting as GitHub workflows, plus small `python-publish.yml`/`version.py` touches.

**Overlap check:** PR #1389 (CI for tests and package build) is *complementary*, not duplicative — #1389 runs tests/build, this adds lint/security. A repo wants both halves eventually.

**But as it stands:** draft for ~2 years, author inactive, and the touched `python-publish.yml`/`version.py` have churned since, so it won't apply cleanly. The workflow configs themselves (`codeql.yml`, `ruff.yml`) are small and mostly evergreen.

**Recommendation:** close as stale, and harvest — the two workflow files could be re-added nearly verbatim on top of #1389's CI once that merges, as a fresh small PR crediting @tamimfatahi. Happy to prepare that follow-up when the CI baseline lands, if wanted. One caution: enabling Ruff on the current codebase will produce a large initial violation list; it should start advisory-only (non-blocking) or scoped to changed files, or it will redline every future PR.

Issue #1313, opened 2024-10-03
Encode NEA Query Properly and Modify Exoplanet Lookup Name


tamimfatahi:
- [x] Encode NEA Query Properly (@ivenzor) 
- [x] Modify Exoplanet Lookup Name (@tamimfatahi)

PR #1314, opened 2024-10-03
Issue 1313: Using Regex to Query Planet Name (Part 2)

tamimfatahi:
This updated edition will query names properly in the locally stored JSON file of exoplanet names as it handles all special characters, leaving only numerics and letters.

Closes Issues #1313


exotic/api/nea.py
```diff
@@ -220,10 +221,13 @@ def _new_scrape(self, filename="eaConf.json"):
         if os.path.exists('pl_names.json'):
             with open("pl_names.json", "r") as f:
                 planets = json.load(f)
-                for key, value in planets.items():
-                    if self.planet.lower().replace(' ', '').replace('-', '') == key:
-                        self.planet = value
-                        break
+                planet_key = re.sub(r'[^a-zA-Z0-9]', '', self.planet.lower())
```
jpl-jengelke, line 224: A comment here would be appreciated. This is basically removing all non-alphanumerics (including spaces) and lowercasing target names. This may be a better approach than used currently in CITISENS, not sure, but I'm also not convinced yet that all target aliases are accounted for here. _Nothing to change, maybe add comment._

jpl-jengelke (approved):

exotic/api/nea.py
jpl-jengelke: @tamimfatahi

Issue #1099, opened 2022-11-07
Do we take into account proper motion?


rzellem:
When we do the plate solution and then check the user's [x,y] position of the target star, do we take into account the proper motion of the host star from the J2000 RA+DEC to the current day?

tamimfatahi:
No, we should do this. I'll scrape it from the NEA and factor it in. Thanks!

Issue #1317, opened 2024-10-09
Modify coordinate entry margin from 100" to 5 PSF


tamimfatahi:
Margin of uncertainty is too large on user-entered coordinates. Would be better to determine the margin using PSF.

PR #1318, opened 2024-10-09
Issues 1099, 1317: Modifying margin from 100" to using PSF for user-entered coordinates and updating coordinates with proper motion

tamimfatahi:
100" may seem too large of a bound for some equipment, which may lead to incorrect user-entered coordinates that EXOTIC would validate. Instead, this PR uses the PSF to help create the bounds.

As talked about before with @rzellem, it has bounds of 1 PSF. However, after testing, it seems as though it is too constraining. I recommend we do something larger, just as we do in our PSF fitting routine where the max goes up to 6.

Closes issue #1099 and issue #1317


exotic/exotic.py
```diff
@@ -713,43 +713,67 @@ def deg_to_pix(exp_ra, exp_dec, ra_list, dec_list):
...
-                return pixx, pixy
+            input_x_pixel, input_y_pixel = prompt_user_for_coordinates(input_x_pixel, input_y_pixel,
+                                                                       calculated_x_pixel, calculated_y_pixel)
+
+
+def validate_pixel_coordinates(input_x_pixel, input_y_pixel, centroid_x, centroid_y, sigma_x, sigma_y, ra_list, dec_list):
+    ra_value = ra_list[int(input_y_pixel)][int(input_x_pixel)]
+    dec_value = dec_list[int(input_y_pixel)][int(input_x_pixel)]
+
+    if not (centroid_x - sigma_x <= ra_value <= centroid_x + sigma_x):
```
tamimfatahi, line 748: Here is the line that specifically uses the PSF to determine the bounds.

tamimfatahi:
Added in the ability to use proper motion to update the target star's coordinates at the time observed.

tamimfatahi:
Just as a heads up - parameters not entered (distance, PM RA, PM DEC) will have a value of 0 if gathering params from the JSON file. Therefore, proper motion will not be accounted for properly.

exotic/plots.py
```diff
@@ -51,7 +51,7 @@ def plot_centroids(x_targ, y_targ, x_ref, y_ref, times, target_name, save, date)
 
 def plot_fov(aper, annulus, sigma, x_targ, y_targ, x_ref, y_ref, image, image_scale, targ_name, save, date, opt_method, min_aper_fov, min_annulus_fov):
     ref_circle, ref_circle_sky = None, None
-    picframe = 5 * (aper + 15 * sigma)
+    picframe = 10 * (aper + 15 * sigma)
```
jpl-jengelke, line 54: Should 15 be a float (e.g. 15. )?

inits.json
```diff
@@ -77,7 +77,10 @@
             "Star Metallicity (-) Uncertainty": -0.08,
             "Star Surface Gravity (log(g))": 4.22,
             "Star Surface Gravity (+) Uncertainty": 0.04,
-            "Star Surface Gravity (-) Uncertainty": -0.04
+            "Star Surface Gravity (-) Uncertainty": -0.04,
+            "Star Distance (pc)": 289.21,
+            "Star Proper Motion RA (mas/yr)": -9.82,
+            "Star Proper Motion DEC (mas/yr)": 3.48
     },
     "optional_info": {
```
jpl-jengelke, line 85: A comment here would not hurt to let readers know HAT-P-32 b is just a default reference. ...

jpl-jengelke (approved):

jpl-jengelke:
Review complete. Please merge upon resolving comments. This must be tested with all merges in `develop`. Thanks.

IraLeeBell:
I've created this test for nea.py. You can run it from the /exotic/api/ folder: https://gist.github.com/IraLeeBell/ac3f7202f29f9b3a3c45e8725dd2e1e7

IraLeeBell:
I'm getting the following error/warning when the _correct_ pixels are selected. I wanted to test this first, before testing the _incorrect_ pixels. I note that EXOTIC is calculating the pixel coordinates at [0,0]. I can confirm that this data set is a valid set, which produces a good light curve and will result in PSF photometry. Also, I can confirm that a plate solution is found when getting this warning.

image

IraLeeBell:
Quick script to determine whether a FIT file has a plate solution in the header (per convo w/ @tamimfatahi ): https://gist.github.com/IraLeeBell/f272dabddec4585c3f9563dde68241c8

inits.json
tamimfatahi: At the top of the file, it states that the initialization file is for HAT-P-32 b.

exotic/plots.py
tamimfatahi: Added.

tamimfatahi:
> I've created this test for nea.py. You can run it from the /exotic/api/ folder: https://gist.github.com/IraLeeBell/ac3f7202f29f9b3a3c45e8725dd2e1e7
> 
> Perhaps it wouldn't hurt to have a test for each of the api files as I've done here (separate from this PR).

Agreed. I'd place the tests in the `tests/` directory such as the one you created. Maybe even add in `pytest` if you can and we can have it run every single time someone opens a PR.

tamimfatahi:
> Quick script to determine whether a FIT file has a plate solution in the header (per convo w/ @tamimfatahi ): https://gist.github.com/IraLeeBell/f272dabddec4585c3f9563dde68241c8

I believe we currently use [astropy](https://docs.astropy.org/en/stable/api/astropy.wcs.WCS.html#astropy.wcs.WCS) to determine whether a FITS file has a plate solution. If that isn't the best way, please let me know.

tamimfatahi:
> I'm getting the following error/warning when the _correct_ pixels are selected. I wanted to test this first, before testing the _incorrect_ pixels. I note that EXOTIC is calculating the pixel coordinates at [0,0]. I can confirm that this data set is a valid set, which produces a good light curve and will result in PSF photometry. Also, I can confirm that a plate solution is found when getting this warning.
> 
> image

For some reason, the archive isn't giving the distance to WTS-2 when querying as its value is 0.0. It does scrape its proper motions though. That's why you're getting such an off value. I updated the code to handle for the error. You should get a warning like this:

>  Warning: Cannot properly account for proper motion due to missing values in: 
Distance (pc). Please re-run and fill in values in initialization file to account for proper motion

IraLeeBell:
Confirming new warning showing when distance in parsecs is set to 0.0 in the inits.json file. 
image

IraLeeBell:
Confirming that when manually entering the parsecs in the inits.json file, we kick off the `check_parameters` `if different` is true process.

image

ivenzor:
I believe EXOTIC use is_celestial from `astropy.wcs import WCS`

`            wcs_file = check_wcs(inputfiles[0], exotic_infoDict['save'], exotic_infoDict['plate_opt'])
`

```
def check_wcs(fits_file, save_directory, plate_opt, rt=False):
    wcs_file = None

    if plate_opt == 'y' and not rt:
        wcs_file = get_wcs(fits_file, save_directory)
    if not wcs_file:
        if search_wcs(fits_file).**is_celestial**:
            log_info("Your FITS files have WCS (World Coordinate System) information in their headers. "
                     "EXOTIC will proceed to use these. "
                     "NOTE: If you do not trust your WCS coordinates, "
                     "please restart EXOTIC after enabling plate solutions via astrometry.net.")
            wcs_file = fits_file

    return wcs_file


def search_wcs(file):
    with warnings.catch_warnings():
        warnings.simplefilter('ignore', category=FITSFixedWarning)
        header = fits.getheader(filename=file)
        return WCS(header)
        # return WCS(fits.open(file)[('SCI', 1)].header)
```

IraLeeBell:
And most importantly, now confirming that when parsecs are manually entered **and** we adopt the parameters and the coordinates are incorrect for the target star, the correct coordinates are suggested. 

image

IraLeeBell (dismissed):
Just ran through a flow with exoplanet WTS-2 b and EXOTIC ran as expected. All files present, looks good.

exotic/exotic.py
```diff
@@ -713,43 +713,78 @@ def deg_to_pix(exp_ra, exp_dec, ra_list, dec_list):
...
+def get_psf_parameters(image_data, x_pixel, y_pixel):
+    psf_data = fit_centroid(image_data, [x_pixel, y_pixel], 0)
+    return psf_data[0], psf_data[1], psf_data[3], psf_data[4]
+
+
+def check_coordinates(input_x_pixel, input_y_pixel, centroid_x, centroid_y, sigma_x, sigma_y,
+                      calculated_x_pixel, calculated_y_pixel):
     while True:
         try:
-            uncert = 1 / 36
```
jpl-jengelke, line 720: Not that this matters (eliminated code) but int divided by int is always float in Python 3. However, such expressions I believe benefit readability by explicitly declaring the nature of the numerics, e.g. 
> 1. / 36. preferred over 1 / 36

exotic/exotic.py
```diff
@@ -713,43 +713,78 @@ def deg_to_pix(exp_ra, exp_dec, ra_list, dec_list):
...
+    """
+    x_min = centroid_x - (sigma_x * 5)
+    x_max = centroid_x + (sigma_x * 5)
+    y_min = centroid_y - (sigma_y * 5)
+    y_max = centroid_y + (sigma_y * 5)
+
+    if not (x_min <= input_x_pixel <= x_max):
+        log_info("\nWarning: The X Pixel Coordinate entered does not match the target's Right Ascension.", warn=True)
+        raise ValueError
+    if not (y_min <= input_y_pixel <= y_max):
```
jpl-jengelke, line 762: Is it OK if they're all the same value?

exotic/exotic.py
```diff
@@ -1108,57 +1143,38 @@ def exp_time_med(exptimes):
...
+def update_coordinates_with_proper_motion(info_dict, time_obs):
+    parameter_names = {
+        'dist': 'Distance (pc)',
+        'pm_ra': 'Proper Motion RA (mas/yr)',
+        'pm_dec': 'Proper Motion DEC (mas/yr)'
+    }
 
-    t = Time(dateobs, format='isot', scale='utc')
-    coordpm = coord.apply_space_motion(new_obstime=t)
+    missing_values = [parameter_names[key] for key in ['dist', 'pm_ra', 'pm_dec'] if info_dict.get(key) == 0.0]
```
jpl-jengelke, line 1153: What if `info_dict.get(key)` returns `None` because the key does not exist? Then that's false and the values a ren't replaced. I would suggest `info_dict.get(key. 0.0)`. ... That is unless there's a desire to potentially not have missing keys recognized.

inits.json
jpl-jengelke: Thanks, not visible in the diff snippet.

exotic/exotic.py
```diff
@@ -1108,57 +1143,38 @@ def exp_time_med(exptimes):
...
+    if missing_values:
+        missing_values = ", ".join(missing_values)
+        log_info("Warning: Cannot account for proper motion due to missing values in: "
+                 f"\n{missing_values}. Please re-run and fill in values in the initialization file to account "
+                 f"for proper motion", warn=True)
 
-    pixcoord = wcs.wcs_world2pix([[coordpm.ra.value, coordpm.dec.value]], 0)
+        return info_dict['ra'], info_dict['dec']
+    else:
+        time_j2000 = Time(2000, format='jyear')
```
jpl-jengelke, line 1163: Minor because it auto-converts it but I see most values entered into `Time()` as floats, e.g. `2000.0`...

jpl-jengelke:
Please take a quick look at the comments and they should be able to be cleared readily.

exotic/exotic.py
tamimfatahi: Technically, yes. If somehow `x_min`, `input_x_pixel`, and `x_max` all contain values of 5, indicating that there is no "spread" in the point-source and that there is only one pixel position that is correct, then I'm okay with it not executing. It's unlikely, but I'm still okay with it occurring.

exotic/exotic.py
tamimfatahi: Makes sense, I left no default value since I ensure just before this step that it would contain the key. Also, we go through most keys in the same dict and assign values of 0.0 if they are `None`.

exotic/exotic.py
tamimfatahi: Done.

exotic/exotic.py
jpl-jengelke: OK, fair enough.

Issue #1254, opened 2024-04-24
pandas.Series.bool is deprecated since Pandas 2.1.0


IraLeeBell:
Please see the following pandas document: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.bool.html

We currently use the following code in nea.py:

`if nans[k].bool():  # if col value is nan`

Need to update to the following for the above example:

`if nans[k].iloc[0]:`

...

Just creating this issue for now and will update/fix after I have a look to see if we're using the Series.bool anywhere else.

IraLeeBell:
If anyone should tackle this before I get to it, please make sure to update the requirements.txt to a modern version of Pandas, such as 2.2.2. Ping me in Slack w/ any questions.

ivenzor:
@IraLeeBell  It seems this is the only part of the code base where pandas.Series.bool is used. Would you like to replace _if nans[k].bool()_ with _if nans[k].any()_ instead of _if nans[k].iloc[0]:_  ?

PR #1323, opened 2024-10-15
General security updates for application dependencies

jpl-jengelke:
- Fix for Python 3.12 incompatibilities
- Upgrade following modules:
> +astroalign~=2.6.0
+astropy~=6.1
+colour_demosaicing~=0.2.6
+matplotlib~=3.9.2
+numpy~=1.26.4
+pandas~=2.2.3
+panel~=1.5.2
+photutils~=2.0.0
+pyvo~=1.5.3
+rebound~=4.4.3
+scipy~=1.14.1

- Tested with `HAT-P-32 b`
- Not yet tested/vetted on Colab


tamimfatahi:
Let's factor in issue #1254 as well. I'll add it onto the branch.

requirements.txt
```diff
@@ -1,27 +1,27 @@
...
 imreg_dft~=2.0.0
 LDTk~=1.8.4
 lmfit~=1.3.2
-matplotlib~=3.7.5
-numpy~=1.24.4
-pandas~=1.5.3
-panel~=1.4.5
-photutils~=1.13.0
+matplotlib~=3.9.2
+numpy~=1.26.4
```
tamimfatahi, line 13: Something worth keeping an eye out on the future is updating to bumpy 2.0 once our other dependents do so.

tamimfatahi (approved):
Tested on a couple of datasets and passed using Python 3.10.

requirements.txt
jpl-jengelke: Most of them are Numpy 2.x compatible now. But we couldn't update Numpy due to an inconsistency with LDTk. So the Numpy version was rolled back to the most recent that has not caused an issue. Numpy 1.26.4 is the last in the 1.x line and released in February 2024. Until LDTk is updated (or we can perhaps update Numba and fix another package hdbscan==0.8.39) any Numpy update will be inconsistent. It will also require deleting some cached HDB files to reset them. 
``` 
ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject```

Bottom line is it's the best we can do for now, I think.

ivenzor:
Tested a dataset on Windows with the new requirements, everything good.

PR #1324, opened 2024-10-15
Release EXOTIC 4.2.0: Plotting enhancements, data precision improvements and API modernization

jpl-jengelke:
- Fix scales on FOV plots and axial range setting 
  - Set circular reticule to red when aperture photometry is selected 
  - Add `opt_method`, `aper` and `annulus` labels to legend 
- Properly scale bound for host star isolation in plate solutions 
  - Account for J2000 RA and DEC 
  - Bound user coordinates for accuracy (limit 5 point spread function[PSF]) 
  - Update conditions for using PSF versus aperture identification
- Stability improvements in querying and analyzing known archive priors 
- Dependency updates to support Python 3.12+
- Various enhancements, bug fixes and stability improvements


exotic/python
rzellem, line 1: ...do we need this empty file?

rzellem (changes requested):
Do we need the empty python file: `exotic/python`?

exotic/python
jpl-jengelke: No, I don't believe we do ... :)

rzellem (approved):

exotic/python
jpl-jengelke: Done in #1325 ...

PR #1328, opened 2024-10-16
Issue 1327: Adding in additional planetary parameter prompts for proper motion values

tamimfatahi:
When checking against the archive, EXOTIC needs additional prompts for the new parameters. My apologies for not catching this earlier. We need to release this for 4.2.1.


exotic/version.py
```diff
@@ -1 +1 @@
-__version__ = '4.2.0'
+__version__ = '4.2.1'
```
jpl-jengelke, line 1: Thanks for updating the version string!

jpl-jengelke (approved):

PR #1329, opened 2024-10-16
Issue 1125: Include AAVSO comparison stars as default during initialization file construction for Colab

### LINKED ISSUE #1125: Colab: add AAVSO comp stars to inits file construction
I think the Colab was made/released before the new AAVSO comp star feature. This should be turned on by default when the inits file is made by the Colab just before EXOTIC is run. Otherwise, right now, EXOTIC will prompt the user if they want to use the AAVSO comp stars or not in the terminal while EXOTIC is running.

tamimfatahi:
I was testing out the colab notebooks earlier and saw that it was prompting AAVSO comp stars from the command line. I also noticed it was an issue asking for the comparison stars to be default per issue #1125. I updated the construct of the file to conform to today's initialization file.

**Please do not release this in v4.2.** Save for 4.3 (next week).

Closes #1125

### DIFF
--- a/exotic/api/colab.py
+++ b/exotic/api/colab.py
@@ -370,10 +370,13 @@ def make_inits_file(planetary_params, image_dir, output_dir, first_image, targ_c
             "Observing Notes": "%s",
 
             "Plate Solution? (y/n)": "y",
-            "Align Images? (y/n)": "y",
+            "Add Comparison Stars from AAVSO? (y/n)": "y",
 
             "Target Star X & Y Pixel": %s,
-            "Comparison Star(s) X & Y Pixel": %s
+            "Comparison Star(s) X & Y Pixel": %s,
+            
+            "Demosaic Format": null,
+            "Demosaic Output": null
     },    
     "optional_info": {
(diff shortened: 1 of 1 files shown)


exotic/api/colab.py
```diff
@@ -370,10 +370,13 @@ def make_inits_file(planetary_params, image_dir, output_dir, first_image, targ_c
             "Observing Notes": "%s",
 
             "Plate Solution? (y/n)": "y",
-            "Align Images? (y/n)": "y",
+            "Add Comparison Stars from AAVSO? (y/n)": "y",
 
             "Target Star X & Y Pixel": %s,
-            "Comparison Star(s) X & Y Pixel": %s
+            "Comparison Star(s) X & Y Pixel": %s,
```
jpl-jengelke, line 376: This is a little clunky since some strings are surrounded with quotes and some are not. I see why and that it works (`null` being subbed into a multiline string) but it creates a disjoint appearance while reading code. 

Just a comment on maintainable code, nothing to strictly fix here.

jpl-jengelke (approved):

PR #1335, opened 2024-10-23
Issue 1125: Temporarily disabling automatic AAVSO Comp Star syncronization for Colab users

tamimfatahi:
Due to scraping AAVSO stars with a given radius from the target star, we need to change the code  to scraping  comp stars from the center of the image instead. I will bring this back in the next release.


jpl-jengelke (approved):
Thanks. This is a temporary change until further coding fully can address comp star synchronization issues.

Issue #1340, opened 2024-11-03
Resolve TypeError for noisy or drifting background imaging


ivenzor:
```
  File "/usr/local/bin/exotic", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.10/dist-packages/exotic/exotic.py", line 2222, in main
    aper_data[f"{ckey}_bg"][i][a][an] = aperPhot(imageData, j + 1, psf_data[ckey][i, 0],
  File "/usr/local/lib/python3.10/dist-packages/exotic/exotic.py", line 1256, in aperPhot
    aperture_sum = (mask.data * (data_cutout - bgflux)).sum()
TypeError: unsupported operand type(s) for -: 'NoneType' and 'float'
```

ivenzor:
The issue appears to happen when comparison stars drift  near the border, but also in other weird cases such as the second one (I added this warning to find more about the issues):

```
Finding transformation 21 of 64 : C:\Users\ivenz\Downloads\9aea8f2e5c67db1b07861583a6ae482e-20241103T223534Z-001\9aea8f2e5c67db1b07861583a6ae482e\MOBS_Cecilia_WASP-107_2021-05-28T19_43_48.971-0700_f142031ed571c5c93969786c1d0cc830_WASP-107210529040312.FITS.gz
Warning: Aperture outside image bounds for star index 6 at position (181.7890170460345, -8.942431356537716)
```

```
Finding transformation 44 of 64 : C:\Users\ivenz\Downloads\9aea8f2e5c67db1b07861583a6ae482e-20241103T223534Z-001\9aea8f2e5c67db1b07861583a6ae482e\MOBS_Cecilia_WASP-107_2021-05-28T19_43_48.971-0700_f142031ed571c5c93969786c1d0cc830_WASP-107210529051212.FITS.gz
Warning: Aperture outside image bounds for star index 4 at position (-7683.0, 2081.5)
```

@tamimfatahi @pearsonkyle 
To avoid these errors, we could check if by any reason data_cutout return None, something like the following:
```
# Method calculates the flux of the star (uses the skybg_phot method to do background sub)
def aperPhot(data, starIndex, xc, yc, r=5, dr=5):
    # Check for invalid coordinates
    if np.isnan(xc) or np.isnan(yc):
        return 0, 0
    
    # Calculate background if dr > 0
    if dr > 0:
[...]

PR #1341, opened 2024-11-05
PR to solve Issue 1340

ivenzor:
Explicitly handle cases where `data_cutout is None` to prevent EXOTIC from crashing due to ` TypeError: unsupported operand type(s) for -: 'NoneType' and 'float'.` 

#1340


ivenzor:
@tamimfatahi

jpl-jengelke (approved):

PR #1343, opened 2024-11-05
Release EXOTIC 4.2.3 Hotfix: Fix TypeError for noisy or drifting background imaging

jpl-jengelke:
- Modify flux calculations to ignore empty data cutouts and workaround with intelligent defaults
- This impacts broad ranging data sets that are either noisy or drifting toward edge of frame


tamimfatahi (approved):
Looks good.

Issue #1346, opened 2024-11-13
Pylightcurve breaking DB upgrade from 4.0.1 to 4.0.4 causing execution failures


ivenzor:
`pylightcurve` was very recently upgraded from` 4.0.1` to `4.0.4`, and the upgrade brings significant changes: https://github.com/ucl-exoplanets/pylightcurve/commit/72bd85783f971b9d60837bb5e7ba3299a19f32be

In particular, it changes how the ExoTETHyS, ephemerides, and photometry databases are cached and used. This can break EXOTIC, even if you’re using virtual environments, if some environments used Pylightcurve 4.0.1 and others are now using 4.0.4. The issue was initially found by Allen and then expanded upon by  @mikeprimm https://uol-ets.slack.com/archives/CJJRP4QKV/p1731417925869009
I'm able to replicate the error in my local installation.

My proposal is as follows: we could pin Pylightcurve to exactly 4.0.1. Currently, we allow pylightcurve~=4.0.1, which will now install 4.0.4 for new installations, causing errors in EXOTIC if an "old" 4.0.1 installation was used previously on that system. This is NOT affecting Colab users as we have a new environment each time. This will likely affect all local users when they upgrade EXOTIC or create a new virtual env with a new EXOTIC installation.

I’m going to submit a PR for your consideration. It would pin the version of pylightcurve to 4.0.1, until we have an easy way to handle the pylightcurve upgrade (Michael mentioned a possible workaround manually flushing the "old" cache) and to review if there are other major changes we need to be aware related to pylightcurve upgrade from 4.0.1 to 4.0.4

@jpl-jengelke @rzellem @tamimfatahi

ivenzor:
PR #1347

mikeprimm:
I was able to get 4.0.4 working, but it seemed to require manual flushing of the locally cached database files.  pylightcurve currently does caching at an unconditional location (that is VERY virtual environment hostile) - basically "<home-dir>/.pylightcurve4" - and I suspect the changes mentioned above were made without dealing with 'cache consistency' (that is, change of format of the databases without invalidating the existing caches), but I'm not 100% sure.  It'd track with being able to get things functioning by zapping the cache directory, though.

ivenzor:
It seems that if you had 4.0.1 on your system and upgraded to 4.0.4, the 'fix' is to manually delete the .pylightcurve4 cache and then let 4.0.4 load the new data.

[...]

PR #1347, opened 2024-11-13
Issue #1346: Workaround for PyLightCurve DB upgrade bug

ivenzor:
Pin pylightcurve==4.0.1 to address issue #1346


requirements.txt
```diff
@@ -14,7 +14,7 @@ numpy~=1.26.4
 pandas~=2.2.3
 panel~=1.5.2
 photutils~=2.0.0
-pylightcurve~=4.0.1
+pylightcurve==4.0.1
```
jpl-jengelke, line 17: A Numpy bug was fixed in 4.0.3. Otherwise, there is some concern about using an almost 4-year-old version. Please examine the comments in #1346 ...

jpl-jengelke (changes requested):

jpl-jengelke:
This is important enough to be part of a HotFix.

requirements.txt
jpl-jengelke: Yes, it looks like this may be the only option right now. ...

jpl-jengelke (approved):

Issue #1348, opened 2024-11-19
Negative aperture values & Negative Pixel Scale in FOV plots


ivenzor:
In exotic a negative aperture is used to mark that no comparison stars is being used. When reporting the aperture in those cases, the negative aperture is converted to positive using abs:
```
            if photometry_info['min_aperture'] == 0:  # psf
                log_info(f"Best Comparison Star: #{photometry_info['comp_star_num']}")
                log_info(f"Minimum Residual Scatter: {round(photometry_info['min_std'] * 100, 4)}%")
                log_info("Optimal Method: PSF photometry")
            elif photometry_info['min_aperture'] < 0:  # no comp star
                log_info("Best Comparison Star: None")
                log_info(f"Minimum Residual Scatter: {round(photometry_info['min_std'] * 100, 4)}%")
                log_info(f"Optimal Aperture: {abs(np.round(photometry_info['min_aperture'], 2))}")
                log_info(f"Optimal Annulus: {np.round(photometry_info['min_annulus'], 2)}")
            else:
                log_info(f"Best Comparison Star: #{photometry_info['comp_star_num']}")
                log_info(f"Minimum Residual Scatter: {round(photometry_info['min_std'] * 100, 4)}%")
                log_info(f"Optimal Aperture: {np.round(photometry_info['min_aperture'], 2)}")
                log_info(f"Optimal Annulus: {np.round(photometry_info['min_annulus'], 2)}")
            log_info("*********************************************\n")
```

Another example:

```
        #######################################################################
        # print final extracted planetary parameters
        #######################################################################

        log_info("\n*********************************************************")
        log_info("FINAL PLANETARY PARAMETERS\n")
        log_info(f"          Mid-Transit Time [BJD_TDB]: {round_to_2(myfit.parameters['tmid'], myfit.errors['tmid'])} +/- {round_to_2(myfit.errors['tmid'])}")
        log_info(f"  Radius Ratio (Planet/Star) [Rp/R*]: {round_to_2(myfit.parameters['rprs'], myfit.errors['rprs'])} +/- {round_to_2(myfit.errors['rprs'])}")
        log_info(f"           Transit depth [(Rp/R*)^2]: {round_to_2(100. * (myfit.parameters['rprs'] ** 2.))} +/- {round_to_2(100. * 2. * myfit.parameters['rprs'] * myfit.errors['rprs'])} [%]")
[...]

PR #1349, opened 2024-11-20
Fix Issue 1348 (Negative Values in FOV plot when no comparison star is used; Negative pixel scale in FOV plot)

ivenzor:
This PR fixes issue #1348 
The plot_fov function was modified to fix the negative values in the FOV plots when no comparison star is used.

image
FOV_Qatar-1 b_2016-06-04_LogStretch


tamimfatahi:
That's a great find. What I am also curious about is why is the scaled pixel values negative too here? They should all be positive.

ivenzor:
@tamimfatahi you are right! I didn't notice that. I was too focused on the negative aperture that I didn't noticed that the pixel scales had negative values too. I will take a look at it later tonight.

ivenzor:
@tamimfatahi about the negative pixel values in the FOV plots, it seems these values are present in the raw image data, not introduced by the visualization pipeline. Raw image statistics show predominantly negative values:

Mean value: -1797.61
Median value: -1807.50
ZSCALE LIMITS:
vmin: -1861.00
vmax: -23.00
**SIMPLE PERCENTILE LIMITS (1-99%):
vmin: -1819.50
vmax: -1717.50**

I guess the negative values likely come from the data reduction/calibration process.

tamimfatahi:
> @tamimfatahi about the negative pixel values in the FOV plots, it seems these values are present in the raw image data, not introduced by the visualization pipeline. Raw image statistics show predominantly negative values:
> 
> Mean value: -1797.61 Median value: -1807.50 ZSCALE LIMITS: vmin: -1861.00 vmax: -23.00 **SIMPLE PERCENTILE LIMITS (1-99%): vmin: -1819.50 vmax: -1717.50**
> 
> I guess the negative values likely come from the data reduction/calibration process.

@ivenzor Thank you for the investigation. I'll accept this PR as this relates specifically to the aperture. However, if you can look into why that's occurring as well, it would be much appreciated.

tamimfatahi (dismissed):

jpl-jengelke:
@ivenzor @tamimfatahi Is there anything else to be done here before it's ready for merge? Specifically, I am asking about the comment on why negative coordinates show up in the FOV plots.

tamimfatahi:
> @ivenzor @tamimfatahi Is there anything else to be done here before it's ready for merge? Specifically, I am asking about the comment on why negative coordinates show up in the FOV plots.

I think @ivenzor  would need to look into why this is occurring. I can't really tell where in the process it's happening.

@ivenzor You mentioned that in the raw image it is also negative? Are you saying that pre-EXOTIC, the images are negative? Or EXOTIC is not calibrating the images properly?

ivenzor:
@tamimfatahi @jpl-jengelke The input images before exotic have positive values as expected:
```
File: MOBS_Cecilia_Qatar-1_2016-06-04T21_01_22.382-0700_9a3fc29eb383d5edaaafb215402400df_Qatar-1160605040122.FITS.gz
--------------------------------------------------
Header Information:
TELESCOP: Cecilia
INSTRUME: Main
EXPTIME: 60.0
DATE-OBS: 2016-06-04T21:01:22.382-0700

RAW IMAGE STATISTICS:
Min value: 215
Max value: 4095
Mean value: 386.43
Median value: 385.00

SIMPLE PERCENTILE LIMITS (1-99%):
vmin: 371.00
vmax: 421.00

Shape: (500, 650)
Data type: >i2

Negative pixels: 0 (0.00%)
```

I think the negative values are introduced in the background subtraction. I need to do more testing and think this through more carefully, but if we’re performing differential photometry between the target star and other stars, aren't we primarily interested in the relative flux differences rather than the actual values themselves?  @rzellem 

We could merge the negative FOV fix, and I’ll take a closer look at the negative pixel values opening a new issue.

ivenzor:
Hi @rzellem, I took a look at the images in the data-upload-aavso channel:

1. The pixel scale values feature was introduced around October 10. After this feature was released, you can occasionally find plots with negative pixel scales.

2. The aperture and annulus values were added to plots starting October 16. After this feature was released, you can occasionally find plots with negative aperture values.

To be honest, I hadn’t noticed either issue, even though they’ve appeared since mid-October. There may have been negative apertures and pixel scales before that date, but since they weren’t plotted, there’s no easy way to verify.

Regarding the negative aperture, I became aware of it two days ago when Anthony pointed out a negative aperture value in a plot by Cledison. The fix in this PR was intended to address that case, which is straightforward and, as far as I know, only affects the plotted value.

As for the negative pixel scale, I wasn’t aware of it until yesterday when Tamim pointed it out while reviewing the PR for the negative aperture issue. I haven’t had the chance to do much testing on this yet, but I noticed that the image passed to the plot function contains many negative values, resulting in the negative scale. I checked the original FITS.gz images, and they have normal positive values, so something must be happening during the calibration or background subtraction process.

I’ll do more testing later tonight when I’m home.

ivenzor:
@tamimfatahi @rzellem I think I now why we got negative values, it is indeed in the calibration section of the code, but is not EXOTIC fault, al least not entirely, but based on the dark frames input data:

DARK FRAME 1 STATISTICS:
Min value: 512
Max value: 4095
**Mean value: 4078.89
Median value: 4095.00**

DARK FRAME 2 STATISTICS:
Min value: 174
Max value: 3141
Mean value: 289.19
Median value: 289.00

GENERAL DARK STATISTICS:
Min value: 343.5
Max value: 3618.0
**Mean value: 2184.04
Median value: 2192.00**

RAW FIRST IMAGE STATISTICS:
Min value: 215
Max value: 4095
Mean value: 386.43
Median value: 385.00

FIRST IMAGE AFTER CALIBRATION STATISTICS:
Min value: -3231.0
Max value: 1905.0
**Mean value: -1797.61
Median value: -1807.50**

The all white dark frame is pulling the median of the general dark off and we are overcorrecting the frames, we end with negative values. This only happens if we happen to have a dark file which is not a dark file...

EXOTIC needs to validate somehow that the dark files are indeed "dark". I will work on it and I will add it to this PR for your consideration.

ivenzor:
@tamimfatahi @rzellem The PR now includes fixes for the original negative aperture value in issue #1348, as well as a fix for the issue with negative values in the science frames caused by overcorrection from a faulty dark frame. EXOTIC now identifies and ignores potentially faulty dark frames.

tamimfatahi:
This needs @rzellem input to okay this, but I do have an alternative method:

What if you instead took the median of the image and checked if any pixel is above 3sigma? That could be an alternative way of handling it. The problem with this is that a single pixel can discard the flat due to cosmic rays.

I'll let Rob ponder either method (or a combination of both!).

ivenzor:
@tamimfatahi, I’m afraid I don’t fully understand your suggestion.

My goal is to discard a bad dark frame so it isn’t included in the 'general dark' used to calibrate the science frames. A bad dark frame in this context refers to a frame that isn’t dark at all—for example, in my case, it was a pure white frame, completely saturated.

What I tried to do is use the maximum pixel value of the dark file as a quick proxy for the saturation value (any hot pixel in the dark frame will yield this value). Then, I compare the median value of the entire dark frame to this saturation value. If the median is greater than 50% of the saturation value, I assume the dark file is not valid (it’s too bright). In my case, the bad dark frame’s median was 100% of the saturation value (a completely white frame). I compared this to other good dark frames I had nearby, whose medians were about 10% of the saturation value. That’s why I considered a 50% cutoff.

tamimfatahi:
> What I'm attempting is to discard a bad dark frame so it isn’t included in the 'general dark' used to calibrate the science frames. A bad dark frame in this context refers to a frame that isn’t dark at all—for example, in my case, it was a pure white frame, completely saturated.

Ah, it's completely saturated. Then the method I mentioned wouldn't work since its for cosmic rays mainly.

Wouldn't you want to use something like `np.iinfo` or `np.finfo` along with `arr.dtype` to figure out the max value of the array? What if the maximum value of the dark is 300 ADU, and the median is 290 ADU, but the array's maximum value is 65,535? Wouldn't this block of code discard the image?

If my analysis isn't correct, please let me know.

ivenzor:
Hi @tamimfatahi thanks for your feedback. My approach  is not based on cosmic rays but rather on the presence of hot pixels in the dark frame.  I'm assuming every dark frame has at least one of those who yield the saturation value I need.

About your suggestion, I think that would give us the theoretical max value of the dtype but I don't know how the instrument would use the dynamic range. Would the theoretical max value will be used if the dark file is somewhat saturated? That's why I chose to use the value of a hot pixel as a saturation proxy.  You have more experience in this, if the theoretical value is reached in practice we can go that route. I will do some testing later tonight.

ivenzor:
At least there is one improvement I can make, instead of comparing the max in each dark frame, I can use all the dark files first to get the max and then follow the same logic: use that max as a proxy for saturation and then compared the mean of each dark file frame to that saturation value as the criteria for rejecting the dark frame.

rzellem:
What's the TL;DR - is this ready to be approved?

tamimfatahi:
> What's the TL;DR - is this ready to be approved?

We need your expertise here @rzellem:
1. Is it sufficient to determine the detector's maximum value by analyzing all dark frames to find the maximum, or by examining a single light frame? I don't have enough experience looking at darks for saturated pixels. I do know they will exist in a light image though.

2. Should we discard a dark frame if the median pixel value exceeds 80% of the frame's maximum pixel value? I agree with the tolerance here.

ivenzor:
@tamimfatahi @rzellem   I have added to the PR checking also a science frame to get the saturation value (in addition to the dark frames).

rzellem:
Sounds like @ivenzor took care of #1 - but I would do a check by looking at the max pixel value according to the bit size of each pixel. That won't necessarily tell if you if saturation is occurring (e.g., for low gains), but it will give you a good idea.

For 2: Well, you could potentially have a camera with terrible dark noise - if that's the case, then you could violate your trigger and remove valid data. What about doing a neighbor check- by comparing a dark to other darks? That will tell you if one (or maybe a few) of them is bad. But a user could confuse darks with bias or flats ... and I'm not sure there's a really good way to do that. One potential way would be to see if the FITS header has info about the dark current for the camera, but I doubt that all cameras have this listed in the header. So I think doing a "how does this one image compare to the others of the same cal type" test is probably the safest thing to do.

ivenzor:
@rzellem, following your last recommendation, what would happen if there’s only one dark file? Or if we have more than one but all of them are saturated or too bright?
In that case if we do the clustering approach we won’t be able to discard them.

ivenzor:
@tamimfatahi  I have updated the PR, it will now use similarity among dark frames to exclude dark frames.

exotic/exotic.py
```diff
@@ -1906,9 +1906,28 @@ def main():
...
+                darksImgList = []
+                for darkFile, darkData, darkMedian in darksMedians:
+                    median_ratio = darkMedian / dMedian
+                    if median_ratio > threshold:
+                        log_info(
+                            f"\nWarning: Skipping suspicious dark frame {darkFile}: "
+                            f"median/overall_median = {median_ratio:.2f}\n",
+                            warn=True
+                        )
+                        continue
```
tamimfatahi, line 1930: Additions look great. One thing I'd like done is to not keep the all the individual calibration photos in memory while processing when we only use the median photo in the end. Can we turn each of these calibration sections (flats, darks, and biases) into functions that will allow for garbage cleanup in the end? This would most certainly prevent future crashes in case users have extremely large images. We should only be returning the `generalDark`, `generalFlat`, and `generalBias`.

tamimfatahi (changes requested):

exotic/exotic.py
```diff
@@ -1327,6 +1327,69 @@ def skybg_phot(data, starIndex, xc, yc, r=10, dr=5, ptol=99, debug=False):
...
+
+def process_bias_frames(bias_files):
+    """Process bias frames and return the master bias."""
+    if not bias_files:
+        return None
+        
+    biases_img_list = []
+    for bias_file in bias_files:
+        bias_data = fits.getdata(bias_file)
+        biases_img_list.append(bias_data)
```
tamimfatahi, line 1370: Change this into list comprehension for readability.

exotic/exotic.py
```diff
@@ -1327,6 +1327,69 @@ def skybg_phot(data, starIndex, xc, yc, r=10, dr=5, ptol=99, debug=False):
...
+
+def process_flat_frames(flat_files, master_bias=None):
+    """Process flat frames and return the normalized master flat."""
+    if not flat_files:
+        return None
+        
+    flats_img_list = []
+    for flat_file in flat_files:
+        flat_data = fits.getdata(flat_file)
+        flats_img_list.append(flat_data)
```
tamimfatahi, line 1382: Change this into list comprehension for readability.

tamimfatahi (changes requested):

exotic/exotic.py
```diff
@@ -1334,12 +1334,8 @@ def process_dark_frames(dark_files):
...
     # e.g. to discard saturated dark files that may negatively affect the master dark used to calibrate the science frames
     # First pass: collect all dark frame medians
-    darks_medians = []
-    for dark_file in dark_files:
-        dark_data = fits.getdata(dark_file)
-        dark_median = np.nanmedian(dark_data)
-        darks_medians.append((dark_file, dark_median))
-        
+    darks_medians = [(dark_file, np.nanmedian(fits.getdata(dark_file))) for dark_file in dark_files]
+
```
tamimfatahi, line 1338: Might be an error here. `dark_file` should really be `fits.getdata(dark_file)`, as you are doing when you get the median here. I didn't mind not making this a list comprehension in the previous review I gave as it may have made it more unreadable. I'll let you decide which way is best (current or old). However, I do want the possible bug fixed.

tamimfatahi: I mean the first `dark_file` in `(dark_file, np.nanmedian(fits.getdata(dark_file)))`

tamimfatahi (changes requested):

tamimfatahi (approved):
Looks great. Thank you for going above and beyond on the issue with the added cleanup @ivenzor!

PR #1354, opened 2025-01-29
Release EXOTIC 4.3.0: Comparison star identification, image frame management and dependency error fixes

jpl-jengelke:
- Fix mesh overlay to not extend beyond image bounds
  - Re-enable AAVSO comparison stars by default
- Improve management of dark, bias and flat imagery
  - Bug fix to restrict negative aperture and filter values (used to flag calibrations) from appearing in FOV plots
  - Enhance plot function to eliminate negative aperture when no comparison star is used
  - Exclude bad dark frame files from general calculations
- Fix for EXOTIC launch crash related to Pylightcurve dependency DB upgrade
- Fix for EXOTIC LDTk class errors relating to Numba on Python 3.11 compatibility that impacts dependency module usage


jpl-jengelke:
@ivenzor

rzellem (approved):

PR #1359, opened 2025-01-31
Issue 1358: Resolve application error when calibration files are nonexistent

ivenzor:
- Conditional to check if calibration files are NoneType first!


jpl-jengelke (approved):
Looks good. Thanks for the quick turnaround!

PR #1366, opened 2025-05-22
Switch to new AAVSO API URLs

mikeprimm:
This is in response to announced change in AAVSO API URLs, due to come into effect on 2025/05/31.  I've tested these with current state of the APIs - the new URLs do appear to be live, at present - but we probably should wait to integrate these until they have completed their update and I can test again with them in their new updated state.

The existing APIs are expected to work for some time - the 'requests' package does handle 301 redirects transparently, and it appears that the updated code is compatible enough to not require additional changes - but AAVSO has indicated that the redirects are temporary.

### DIFF
--- a/docs/regions/English/How-to-select-comparision-stars-using-AAVSO-Star-Charts.txt
+++ b/docs/regions/English/How-to-select-comparision-stars-using-AAVSO-Star-Charts.txt
@@ -2,15 +2,15 @@ AAVSO Comparison star charts
 
 1. Getting the charts
 
-   the main URL is https://app.aavso.org/vsp/
+   the main URL is https://apps.aavso.org/vsp/
 
  you can go directly to the chart for a specific star by looking up the chart id in the lists that are periodically posted to the Slack channel and forming the URL as
 
-       https://app.aavso.org/vsp/chart/?chartid=(the listed chartid)
+       https://apps.aavso.org/vsp/chart/?chartid=(the listed chartid)
 
 Example, fot HAT-P-23
 
-      https://app.aavso.org/vsp/chart/?chartid=X26350XQ
+      https://apps.aavso.org/vsp/chart/?chartid=X26350XQ
 
 You can also go to the main URL and use the full search capabilities. You can search by star name or by RA and Dec. There is a help available (https://www.aavso.org/sites/default/files/VSP_Help.pdf)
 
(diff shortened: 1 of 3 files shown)


mikeprimm:
Confirmed with AAVSO that new URLs are considered 'live' and that moving to them ASAP is advisable.  Also, redirects for old URLs should be functional for at least 3 months, so it should be some time before folks running existing code will break - but it'd be advisable to announce that at the appropriate time.

jpl-jengelke:
Obviously the redirects have been active for a bit longer than 3 months... :)

opusbuilds:
Executed review, 2026-08-21 (part of the open-PR triage Rob asked me to run; comments and evidence only, merge decisions stay with maintainers).

**What it does:** swaps three AAVSO URL families to the hostnames AAVSO announced for the 2025-05-31 migration: `www.aavso.org/vsx/index.php` → `vsx.aavso.org/index.php` (VSX API, 2 call sites), `www.aavso.org/apps/vsp/api/chart/` → `apps.aavso.org/vsp/api/chart/` (VSP chart API, 1 call site), and `app.aavso.org/vsp/` → `apps.aavso.org/vsp/` (docs + GUI labels).

**Live test today** (HAT-P-28 field, RA 13.002, Dec +34.728):

| URL | status | redirects | payload |
|---|---|---|---|
| VSX old (`www.aavso.org/vsx`) | 200 | 1 | valid JSON, 1 object |
| VSX new (`vsx.aavso.org`) | 200 | 0 | valid JSON, 1 object |
| VSP old (`www.aavso.org/apps/vsp/api/chart`) | 200 | 2 | valid JSON, chartid X42961AKB |
| VSP new (`apps.aavso.org/vsp/api/chart`) | 200 | 0 | valid JSON, chartid **X42961AKB** (identical) |

So the PR is correct and still needed: 15 months after the migration date the old URLs only survive on redirect chains AAVSO described as temporary, and every EXOTIC run currently pays 1–2 extra hops per query while betting the redirects stay up.

**One gap if merged as-is:** the active development branch (`michael_fitzgeralds_wonderful_branch_of_magic`) rewrites `exotic.py` wholesale and still carries the OLD urls in **four** call sites — the three this PR fixes, plus a new one this PR predates: `vsx_field_query` (line ~15421, the fortuitous-variable field query). Merging to `develop` alone gets silently reverted whenever that branch lands.

**Recommendation:** merge here, and port the same one-line swaps (all four sites) onto the magic branch, crediting @mikeprimm. Happy to open that port PR if wanted — it's a 4-line diff and I've already located the sites.

PR #1372, opened 2026-03-26
Fix issue 1255

### LINKED ISSUE #1255: Fall back to RA+DEC on Colab for FOV Plot

S-Bhattacharya240611:
This PR resolves Issue #1255 by implementing a coordinate-based fallback for the Field of View (FOV) plot.
The Problem:
When the NASAExoplanetArchive query fails (due to target omission, candidate status, or API instability), the pipeline currently fails to generate the FOV plot because of missing coordinates in pDict.
The Solution:
I have implemented a logic bridge in exotic/exotic.py that checks for user-provided RA and DEC values in userpDict (from inits.json) if the NASA Archive lookup returns empty.
Safety Features:
Utilizes .get() for safe dictionary access.
Includes a secondary check to prevent NoneType crashes if both the Archive and the user-input are missing.
Added logging to notify the user when a fallback is active.
Testing:
Verified syntax via py_compile.
Passed all 45 existing unit tests (test_ld.py and test_utils.py) using pytest on Python 3.12.
Resolves #1255

### DIFF
--- a/exotic/exotic.py
+++ b/exotic/exotic.py
@@ -1943,6 +1943,10 @@ def main():
         if not args.override:
             nea_obj = NASAExoplanetArchive(planet=userpDict['pName'])
             userpDict['pName'], CandidatePlanetBool, pDict = nea_obj.planet_info()
+            if not pDict.get('ra') or not pDict.get('dec'):
+                log_info("NASA Archive coords missing. Falling back to inits.json RA/DEC.")
+                pDict['ra'] = userpDict.get('ra')
+                pDict['dec'] = userpDict.get('dec')
         else:
             pDict = userpDict
             CandidatePlanetBool = False
--- a/exotic/exotic.py
+++ b/exotic/exotic.py
@@ -1943,8 +1943,11 @@ def main():
         if not args.override:
             nea_obj = NASAExoplanetArchive(planet=userpDict['pName'])
             userpDict['pName'], CandidatePlanetBool, pDict = nea_obj.planet_info()
-            # --- Fallback to user-provided RA/Dec from inits.json if NASA Exoplanet Archive
-            # returns empty coordinates (e.g., for candidate planets or API failures). ---
+            
+            # --- V2 SOTA FALLBACK (FIXED FOR CANDIDATES) ---
+            if pDict is None:
+                pDict = {}
+
             if not pDict.get('ra') or not pDict.get('dec'):
                 user_ra = userpDict.get('ra')
                 user_dec = userpDict.get('dec')
@@ -1954,8 +1957,8 @@ def main():
                     pDict['ra'] = user_ra
                     pDict['dec'] = user_dec
                 else:
-                    log_info("WARNING: NASA Archive coords missing AND no RA/DEC provided in inits.json. FOV plot may fail.")
-            # ---------------------------------
+                    log_info("WARNING: NASA Archive coords missing AND no RA/DEC provided. FOV plot may fail.")
+            # -----------------------------------------------
         else:
             pDict = userpDict
             CandidatePlanetBool = False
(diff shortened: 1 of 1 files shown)


rzellem:
Have you tested to see if this works as intended if a user does not supply an inits.json file?

S-Bhattacharya240611:
> Have you tested to see if this works as intended if a user does not supply an inits.json file?

Hi Dr. Zellem , I have performed a regression test on the interactive (non-JSON) workflow to verify the fix.
During testing, I identified that planet_info() returns None for candidate planets in interactive mode, which originally caused an AttributeError in my fallback logic. I have updated the implementation to:
Safely initialize pDict as an empty dictionary if the archive returns None.
Successfully perform the RA/DEC fallback from userpDict for both JSON and manual entries.
Verified the fix by passing the full pytest suite (45/45) on Python 3.12.
This ensures the FOV plot is now robust across all input methods.And as always, thank you for the guidance!
The issue is fixed and tested

opusbuilds:

**What it does:** when the NASA Exoplanet Archive query returns no usable coordinates (candidate targets, omissions, API failures), falls back to the inits.json RA/DEC so the FOV plot can still be generated (issue #1255).

**Static assessment:** the direction is right and the issue is real. Three things before merge:

1. **Sparse-pDict scope** — after the patch, a failed archive query yields `pDict` containing *only* `ra`/`dec`. That rescues the FOV plot, but everything downstream that expects period/priors still sees an empty dict. Was this tested end-to-end on a real candidate target (the #1255 scenario), and does the run complete rather than fail later? A one-line answer with the target name you tested would settle it.
2. **Comment hygiene** — `# --- V2 SOTA FALLBACK (FIXED FOR CANDIDATES) ---` describes the patch's history, not the code. Suggest replacing with a comment saying *why* (archive can omit candidates; FOV plot needs coords from somewhere).
3. **The magic branch (#1373) has rebuilt this exact region** — it now guards `isinstance(pDict, dict)`, keeps an `archive_planet_dict`, and does archive↔inits coordinate recovery in both directions (`exotic.py` ~32479–32530). This patch won't apply there, and the #1255 failure may already be fixed on that branch. Worth one test of the #1255 scenario against #1373 before deciding whether this merges to develop as a stopgap or closes as superseded.

**Recommendation:** iterate (answer 1, clean 2), and run the scenario on #1373 — if the magic branch already handles it, close as superseded with the issue cross-referenced; if not, this is a reasonable develop-side stopgap.

opusbuilds:
Ran the #1255 scenario on the magic branch tip (e024049), as promised in the walkthrough.

**Setup:** a real 87-frame MicroObservatory night reduced under the name `TOI-99999.01` (the Exoplanet Archive does not know it; the decimal suffix makes EXOTIC treat it as a candidate), coordinates and priors from the inits file only, `Add Comparison Stars from AAVSO? y`, quick-look mode.

**Result:** exit 0, no traceback. The log shows `Cannot find target (TOI-99999.01) in NASA Exoplanet Archive. Assuming TOI-99999.01 is a planet candidate`, then the full reduction runs and all four FOV plots (`FOV_TOI-99999.01_*Stretch_*`) are written. Reason it works without this patch: on that branch the inits path already does `pDict = userpDict` for candidates (`exotic.py` ~35239-35260), so the inits RA/Dec reach the FOV code; and the AAVSO chart request (`vsp_query`) builds its URL from the WCS image-centre coordinates, never from the target name, so the 2024 name-lookup failure in #1255 cannot recur there. (The chart step is gated on a plate-solved WCS file, which these frames lack, so it was skipped in this run rather than exercised.)

Not tested: the interactive no-inits path Rob asked about. On the magic branch that goes through `get_planetary_parameters(CandidatePlanetBool, ...)`, which prompts for the candidate's parameters.

**Recommendation:** close as superseded by #1373, and add `Closes #1255` to #1373 so the issue closes with the release. If a develop-side stopgap is still wanted before then, the two review points above (sparse pDict; the history comment) still apply.

PR #1373, opened 2026-03-29
Large update for Speed, Robustness and Pre-reduced focussed updates. 

mfitzasp:
Ok, this is a big one. Kalee has tested tens of transits both original and this version. I've used it for tens and theres been a few other users who have used it.... only found a couple of bugs over time. I will be around to catch any more that may arise... I will be using EXOTIC a lot from now... and by "I", I means hordes of students.

Unfortunately I was working on it locally and my commits are quite large! But I am pretty dang sure they are fine at this stage. My major motivation was the make it more robust (fallback options for brittle online services) and faster and actually fixes up a few issues in the pre-reduced area in particular that was less well serviced over time. I found a few clangers in there and there are a bunch of feature-unchanging speedups (e.g. using bottleneck rather than numpy for nan-aware array operations and stopping analysis when the target star is outside the field of view... previously it still analysed the whole frame even though it was futile which is one reason why some runs blew out to 2 hours).

Heres the broad list of changes. I found a couple of small bugs later on as well. 

Major new external-service architecture:

Added NextAstro plate-solving implementation (new source extraction, brightest-source capping, async polling, WCS writeout).

Added robust fallback routing both directions: NextAstro -> nova and nova -> NextAstro.

Added RA/Dec and pixel-scale astrometry hints to uploads.

Added NextAstro variability API integration and JSON request/response logging.

Added fallback from VSX query failures to NextAstro variability checks.

Added NASA Exoplanet Archive fallback via NextAstro cached exoplanet parameters.

Runtime/CLI feature additions:

New flags for NextAstro astrometry, NextAstro variability server, non-interactive mismatch handling, multiprocessing transformations, multiprocessing candidate fits.

Better non-interactive coordinate mismatch behavior (auto-proceeds with WCS-derived coords).

FITS/WCS and transformation pipeline improvements:

Added optional header-WCS ignore override (force legacy manual alignment path).

Added WCS coverage precheck to skip unnecessary transformation precompute when all files have celestial WCS.

Added transformation fast path and staged fallbacks (FFT-first plus optimized fallback behavior).

Added multiprocess precompute with worker cap (8) and clearer progress diagnostics.

Reduced warning spam and improved plate-status handling.

Photometry and model-fitting quality/performance:

Added fast centroiding path with periodic full-fit cadence.

Improved out-of-frame handling and fail-fast behavior for projected stars outside frame.

Added coarse+refined aperture-grid auto-tuning.

[...]

PR #1382, opened 2026-05-24
Fix #1369

### LINKED ISSUE #1369: Incorrect mObs values used by Exotic in Google Colab
In Step 3 of the Exotic Standard Google Colab, the function `make_inits_file` is called. This function is defined in `EXOTIC/exotic/api/colab.py`
That function calls another function, `find`, where a set of default values for mObs is hardcoded:
`  mobs = {"FILTER": "V", "LATITUDE": "+37.04", "LONGITUD": "-110.73", "HEIGHT": 2606 }`
`"Pixel Binning": "1x1",`

However, in` EXOTIC/exotic.py`, there is a different set of mObs values:
```
            # checks for MOBS data
            if 'CREATOR' in header:
                if 'MicroObservatory' in header['CREATOR'] and 'MOBS' not in exotic_infoDict['second_obs'].upper():
                    if exotic_infoDict['second_obs'].upper() != "":
                        exotic_infoDict['second_obs'] += ",MOBS"
                    else:
                        exotic_infoDict['second_obs'] = "MOBS"
                    exotic_infoDict['filter'] = "MObs CV"
                    exotic_infoDict['elev'] = 1268
                    exotic_infoDict['lat'] = 31.675467
                    exotic_infoDict['long'] = -110.951376
                    exotic_infoDict['pixel_bin'] = "2x2"
```

To summarize, there are two different sets of values being used for mObs:
Filter: **MObs CV** (which is translated to just CV in EXOTIC/exotic/api/filters.py) vs **V**
Elevation/Height: **1268** vs **2606**
Latitude: **31.675467** vs **37.04**
Longitude: **-110.951376** vs **-110.73**
Binning: "**2x2"** vs **"1x1"**

I spoke with Frank S., and he confirmed that the correct values should be:
Filter: **CV**
Elevation/Height: **1268**
Latitude: **+31.675467**
Longitude: **-110.951376**
Binning:  **"2x2"**

Based on this, it looks like the values currently used in colab.py are incorrect, and should be updated to match the set in exotic.py.

@rzellem — would you like me to go ahead and create a PR for your review and consideration?

S-Bhattacharya240611:
Auto-fix

### DIFF
--- a/exotic/api/colab.py
+++ b/exotic/api/colab.py
@@ -220,7 +220,7 @@ def convert_Mobs_to_utc(datestamp, latitude, longitude, height):
 def find (hdr, ks, obs):
   # Special stuff for MObs and Boyce-Astro Observatories
   boyce = {"FILTER": "ip", "LATITUDE": "+32.6135", "LONGITUD": "-116.3334", "HEIGHT": 1405 }
-  mobs = {"FILTER": "V", "LATITUDE": "+37.04", "LONGITUD": "-110.73", "HEIGHT": 2606 }
+  mobs = {"FILTER": "CV", "LATITUDE": "+31.675467", "LONGITUD": "-110.951376", "HEIGHT": 1268}
 
   if "OBSERVAT" in hdr.keys() and hdr["OBSERVAT"] == 'Whipple Observatory':
     obs = "MObs"
@@ -365,7 +365,7 @@ def make_inits_file(planetary_params, image_dir, output_dir, first_image, targ_c
             "Obs. Longitude": "%s",
             "Obs. Elevation (meters)": %d,
             "Camera Type (CCD or DSLR)": "CCD",
-            "Pixel Binning": "1x1",
+            "Pixel Binning": "2x2",
             "Filter Name (aavso.org/filters)": "%s",
             "Observing Notes": "%s",
 
--- a/tests/test_colab.py
+++ b/tests/test_colab.py
@@ -1,0 +1,71 @@
+import importlib
+import json
+import sys
+import types
+
+import numpy as np
+from astropy.io import fits
+
+
+def import_colab(monkeypatch):
+    fake_barycorrpy = types.ModuleType("barycorrpy")
+    fake_barycorrpy.utc_tdb = object()
+    fake_ipython = types.ModuleType("IPython")
+    fake_display = types.ModuleType("IPython.display")
+    fake_display.display = lambda *args, **kwargs: None
+    fake_display.HTML = lambda value: value
+
+    monkeypatch.setitem(sys.modules, "barycorrpy", fake_barycorrpy)
+    monkeypatch.setitem(sys.modules, "IPython", fake_ipython)
+    monkeypatch.setitem(sys.modules, "IPython.display", fake_display)
+
+    return importlib.import_module("exotic.api.colab")
+
+
+def test_mobs_defaults_match_exotic_metadata(monkeypatch, tmp_path):
+    colab = import_colab(monkeypatch)
+    fits_path = tmp_path / "mobs.fits"
+    fits.PrimaryHDU(data=np.zeros((2, 2))).writeto(fits_path)
+
+    with fits.open(fits_path, mode="update") as hdul:
+        hdr = hdul[0].header
+        hdr["OBSERVAT"] = "Whipple Observatory"
+        hdr["DATE"] = "2017-12-20T01:33:43.000"
+        hdr["WEATHER"] = 80
+        hdr["TELTEMP"] = 20.0
+        hdr["CAMTEMP"] = 10.0
+
+    hdr = fits.getheader(fits_path)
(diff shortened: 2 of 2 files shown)


S-Bhattacharya240611:
Additional validation after opening PR:

- Re-read README.md and docs/README.md. README recommends validating with the HAT-P-32 b sample dataset.
- Cloned rzellem/EXOTIC_sampledata locally; confirmed 142 FITS files in sample-data/HatP32Dec202017.
- Re-ran python -c "import exotic": passed.
- Re-ran pytest tests/test_utils.py: 32 passed.
- Ran focused Colab validation against the first sample FITS file (OBSERVAT=Whipple Observatory) through exotic.api.colab.find() and make_inits_file(). Verified generated inits values:
  - Filter Name: CV
  - Obs. Latitude: +31.675467
  - Obs. Longitude: -110.951376
  - Obs. Elevation: 1268
  - Pixel Binning: 2x2
  - Secondary Observer Codes: MOBS

Full README sample reduction was also attempted with python exotic/exotic.py -red sample-data/inits-local.json -ov. It started the Complete Reduction routine but failed outside this change while LDTK attempted FTP limb-darkening grid access: ftplib.error_reply: 229 Extended Passive Mode Entered. That failure is in external LDTK data retrieval, not the Colab MObs metadata path changed by this PR.

opusbuilds:

**What it does:** corrects the hardcoded MObs fallback metadata in `colab.py` (filter V→CV, lat/long, elevation 2606→1268) and the inits default binning (1x1→2x2), with a regression test.

**Independent verification against a real MObs frame** (telescope Cecilia, Whipple Observatory, 2026-08-16, from my own reduction archive) — header says:

```
LATITUDE = 31.68      LONGITUD = -110.88     HEIGHT = 1268.0
XBINNING = 2          YBINNING = 2           FILTER = 'Clear'
```

Verdict on each change:
- **Elevation 1268**: exact match. The old 2606 m is the Mt. Hopkins *summit* (MMT), not the MObs site.
- **Binning 2x2**: exact match. Old 1x1 was wrong for MObs Main.
- **Filter CV**: correct — MObs shoots `Clear`, and CV is the AAVSO code for clear-with-V-zeropoint. The old plain `V` wrongly claimed a Johnson V filter.
- **Lat/long**: the old (+37.04, −110.73) is ~600 km from Whipple — plainly wrong. The PR's (31.675467, −110.951376) is the basecamp and is *internally consistent* with 1268 m; interestingly the header's own rounded (31.68, −110.88) pairs the ridge longitude with the basecamp elevation, so the PR is arguably more self-consistent than the headers. Either way the ~7 km difference shifts barycentric time corrections by microseconds — scientifically negligible.

**Test:** `1 passed` in a full environment. Note for the CI work in #1389: it fails in a bare env only because `colab.py` imports `bokeh` at module level — requirements-dev will need the full stack, or the test wants an `importorskip`.

**Conflict:** the magic branch (#1373) carries the same wrong constants (`colab.py:239`, binning at `:385`) — same merge-and-port pattern as #1366.

**Recommendation:** merge; port the same constants to #1373. One open question for @S-Bhattacharya240611: `make_inits_file`'s template also serves the Boyce path — is 2x2 correct for Boyce-Astro frames, or should the binning default be per-observatory?

opusbuilds:
One addition from a re-check today (2026-09-04): the same MObs constants exist in a second copy that this PR does not touch, `exotic/utils.py` `find()`:

```python
mobs = {"LATITUDE": "+37.04", "LONGITUD": "-110.73", "HEIGHT": 2606}
...
if "OBSERVAT" in hdr.keys() and hdr["OBSERVAT"] == 'Whipple Observatory':
    obs = "MObs"
...
if obs == "MObs":
    mobs_val = get_val(mobs, ks)
    if mobs_val:
        return mobs_val
val = get_val(hdr, ks)
```

That is the command-line/inits path (blank latitude/longitude/elevation in the inits file), and it returns the hardcoded values *before* reading the header, so a real MicroObservatory header carrying `LATITUDE 31.68 / LONGITUD -110.88 / HEIGHT 1268.0` (checked on a 2026-08-16 Cecilia frame) is overridden by the wrong constants. The Colab fix here is correct; the utils.py copy needs the same edit or, simpler, should defer to the header when the header has the keys. Both copies are also present on `michael_fitzgeralds_wonderful_branch_of_magic` (`exotic/api/colab.py:239`, `exotic/utils.py:596`); I will port both there together.

Scale of the effect for users, so nobody panics: the barycentric-time error from a ~600 km site offset is milliseconds; the airmass curve is off by several percent, which the detrending mostly absorbs. Worth fixing, does not invalidate anyone's data.

opusbuilds:
Ported to the magic branch in #1410, together with the second copy of the same constants in `exotic/utils.py` `find()` (the command-line path, where they override the header). Pixel binning is read from the frame header there, which should cover the Boyce-Astro case. You're credited as co-author on the port commit.

PR #1389, opened 2026-07-28
Add CI workflow for tests and package build

haggbart:
- **test** - install requirements, run pytest
- **build** - mirrors python-publish.yml (wheel, sdist, twine check)

CI is red (for me) with 5 failures, the suite has an isolation problem: test_exotic_proper_motion.py passes alone (221 tests), but run test_ephemeris_validation.py first and 4 fail on KeyError: 'u0'. The leak's in exotic/exotic.py so I left it.

Also skipped that file's two Windows tests, they assert Windows path semantics with no platform guard.

892 passed, 2 skipped, 5 failed locally (macOS/3.12; CI is ubuntu/3.10).


opusbuilds:

**Transparency flag first:** this PR's author is my collaborator (he administers the server I run on). Treat my review as evidence-only with extra weight on your own judgment.

**What it does:** adds a minimal CI workflow (test job: py3.10, pip cache, `pytest`; build job mirroring python-publish: wheel + sdist + `twine check --strict`), local-test instructions in CONTRIBUTING, `testpaths` config, `requirements-dev.txt` (pytest), and correct `skipif` guards on two Windows-only tests. Security posture is right: `permissions: contents: read`, `persist-credentials: false`, concurrency cancel.

**Executed evidence** (ran exactly what the CI would run, locally):

| run | result |
|---|---|
| full suite, PR head | 5 failed, 892 passed, **2 skipped** (the new skipifs work) |
| full suite, magic-branch tip | 7 failed, 1027 passed |
| each failing test standalone | **passes** |
| each failing test within its own file | **passes** (test_output_files.py alone: 46/46; test_exotic_proper_motion.py alone: only the 2 Windows tests fail without the skipifs) |

Reading of that table: this PR fixes 2 of the tip's 7 whole-suite failures (the Windows ones). The remaining 5 are **pre-existing cross-file test pollution** on the branch — some earlier test file leaves global state that breaks `finalize_comparison_*`, `fit_ranked_*`, and `prior_depth_*` only in whole-suite order. Not introduced by this PR, but it means the shiny new CI would be red on day one, which trains everyone to ignore it.

**Recommendations:**
1. Rebase — the branch has moved 20 commits; this PR's files barely intersect, so it should be clean.
2. Fix the 5-test isolation before or with the merge, so CI starts green. I'm happy to bisect the polluting file pair and post findings as a follow-up — it's a bounded search.
3. Keep the scope this minimal. Version matrix, lint (see #1307 harvest), and timeouts can come later; a green baseline is the valuable thing.

**Bottom line:** highest-leverage PR on the board — every other review gets cheaper once tests run on PRs automatically. Iterate (rebase + green baseline), then merge.

opusbuilds:
Follow-up on the 5 order-dependent failures, as offered — bisection results.

**Minimal reproduction** (fails identically on branch tip, no CI needed):
```
pytest tests/test_boolean_config.py tests/test_exotic_proper_motion.py
```
→ the three `finalize_comparison_*` + `fit_ranked_*` tests fail (plus the two Windows ones #1389 already guards).

**Proximate cause**, same for every victim:
```
exotic/exotic.py:3375: in estimate_transit_duration_samples_from_fit
exotic/api/elca.py:160: KeyError: 'u0'    (_instantaneous_transit reading values['u0'])
```
The victims' fit `values` dict is missing its limb-darkening keys when — and only when — certain other tests ran first.

**Shape of the pollution — it's two-factor, which is why it looks spooky:**
- Any one of **three unrelated files** acts as a trigger: `test_aperture_tuning_performance.py`, `test_boolean_config.py`, or `test_ephemeris_validation.py` (each + the victim file → 6 failures; `test_centroid_wcs.py` and `test_elca_baseline.py` do not trigger).
- But a trigger file + a victim *test alone* passes — the recipe also needs earlier tests **within** `test_exotic_proper_motion.py` to run. So: external trigger + in-file accomplice → victims fail.
- Three unrelated triggers producing an identical KeyError points at shared machinery (module-level state in `exotic.exotic`/`elca` whose initialization is import-order- or call-order-dependent), not at one test forgetting to undo a monkeypatch.

**Suggested next hop** (yours or mine, happy either way): instrument where the `values` dict gets its `u0..u3` — if there's a module-level prior/values template that some code path builds without LD keys once a flag or cache is set, the trigger files are just the ones that flip it first. Until fixed, `pytest -p randomly` would surface this class loudly rather than intermittently.

With this repro in hand the fix is likely small, and #1389's CI can start green.

haggbart:
Rebased, and confirmed your repro locally. One extra data point for the two-factor read: a trigger file plus only those four tests (by node ID) passes – the failures also need earlier tests in test_exotic_proper_motion.py to run first.

opusbuilds:
Heads-up for this PR, 2026-08-25: the magic branch gained its own CI overnight — `python-tests.yml` in 491967a (py3.12+3.13 matrix, `pip check`, MPI-absence guard, full pytest). That covers this PR's **test job** and more, so a rebase now hits a filename/scope collision.

What #1389 still adds that the new workflow lacks: the **build job** (wheel + sdist + `twine check --strict`), pip caching, minimal `permissions:`/`persist-credentials: false` hardening, concurrency-cancel, and the CONTRIBUTING instructions. Suggested reshape rather than close: drop the test job, keep the build job (possibly merged into `python-tests.yml` as a second job), and carry over the hardening flags to Michael's workflow — his currently runs with default permissions on both push and PR.

Also relevant: the whole-suite pollution finding above (KeyError `'u0'`) predates the numpy-2 migration; whether it survives on the new stack is worth one fresh full-suite run before deciding the green-baseline question — the failure was environmental-order, not numpy-related, so it likely persists.

haggbart:
Closing, the branch has its own CI now (#1373) and the numpy 2 migration fixed the failing tests (1042 passed on current tip without this PR).

Two leftovers: python-tests.yml runs with default token permissions, and requirements.txt still has the stray python_version line, now as python-version>=0.0.2.

PR #1391, opened 2026-07-30
Fix silently wrong BJD_TDB when -ov and -nea are combined

### LINKED ISSUE #1390: WBoM: -ov -nea leaves RA/Dec as sexagesimal strings — BJD_TDB silently wrong by 90-540 s (Tmid-scale)
**Branch:** `michael_fitzgeralds_wonderful_branch_of_magic` (tip cb551045). Found during a module audit of the time-conversion chain; every number below is from executed code, not inspection. PR with the fix and a regression test: incoming reference.

## Symptom

When `--override` and `--nasaexoarch` are combined (`-ov -nea`, common for scripted runs), all reported times — Tmid included — are systematically wrong by **~90–540 s depending only on the target's RA**, roughly constant within a night. Measured against a correct barycentric conversion (MicroObservatory site):

| Target | BJD_TDB error | vs typical Tmid error bar |
|---|---|---|
| CoRoT-2 b | **−538 s** | several σ |
| HAT-P-27 b | **−206 s** | ~1σ |
| TOI-1516 b | **−93 s** | inside 1σ |

The log tell for an affected run is this warning, easy to read past as benign:

```
barycorrpy JDUTC_to_BJDTDB conversion failed; falling back to astropy light-travel-time conversion for this run.
```

## Cause chain

1. With `--override`, `pDict = userpDict` — RA/Dec are the init file's **sexagesimal strings** (`"19:27:06.50"`).
2. The conversion dispatch reads `if args.nasaexoarch: pass / elif args.override: radec_hours_to_degree(...)`. With both flags, `nasaexoarch` wins the branch and **the conversion never runs** (`-ov` alone was fixed by 98f015b; the combination was not).
3. `convert_jd_to_bjd` calls barycorrpy with the string RA → `TypeError: can't multiply sequence by non-int of type 'float'` (`barycorrpy/utc_tdb.py:373`).
4. The astropy fallback then does `SkyCoord(p_dict['ra'], p_dict['dec'], unit=(u.deg, u.deg))` — parsing `"19:27:06.50"` as **19.45 degrees instead of 19h27m = 291.78 degrees**. Dec survives (sexagesimal dec really is degrees). The barycentric correction is computed for a point up to ~150° away on the sky, and BJD_TDB shifts by the table above.
5. The same unconverted string reaches the airmass code; with MObs headers this is dormant (TELALT present), but the astropy `SkyCoord(f"{ra} {dec}", unit=deg)` branch gives airmass 1.37 instead of 1.27 when it fires.

Note the fallback itself is correct — fed numeric degrees it agrees with correctly-called barycorrpy to ≤4e-5 s (topocentric, ICRS, TDB scale composition all verified). The bug is purely that strings reach it, and that it parses them plausibly instead of failing.

## Minimal repro

```python
from exotic.exotic import convert_jd_to_bjd
convert_jd_to_bjd([2461222.65], {'ra': '19:27:06.50', 'dec': '+01:23:01.5'},
                  {'lat': 31.68, 'long': -110.88, 'elev': 1268.0})
# logs the barycorrpy TypeError, returns 2461222.64981753
# correct value (numeric degrees, same site): 2461222.65604706  ->  -538.2 s
```

## Fix (in the PR)

Two small changes:
- the dispatch becomes `if args.nasaexoarch and not args.override: pass`, so the combination converts the user's coordinates exactly once (`-nea` alone and `-ov` alone unchanged);
- `convert_jd_to_bjd` defensively coerces string coordinates through `radec_hours_to_degree` before any conversion, so a future path delivering strings converts correctly or fails loudly rather than silently shifting the time base.

With the fix, string and numeric inputs agree to 0.000000 s and barycorrpy handles the conversion (no fallback). Regression test included, asserting both the string/numeric agreement and the absolute offset (+522.5 s for the CoRoT-2 geometry) so a sign regression can't sneak back.

## Who should check their numbers

Anyone scripting WBoM with `-ov -nea` whose logs show the barycorrpy fallback warning. My own three reductions above are affected and I'm re-running them; times produced through the interactive flow or with either flag alone are fine.

opusbuilds:
Fixes #1390.

Root fix: the coordinate-conversion dispatch now runs `radec_hours_to_degree` when both flags are set (`-nea` alone and `-ov` alone unchanged). Defensive fix: `convert_jd_to_bjd` coerces string coordinates before any conversion, so any future path delivering strings converts correctly or fails loudly instead of silently mis-parsing hours as degrees in the astropy fallback.

Verified: string and numeric inputs now agree to 0.000000 s, barycorrpy handles the conversion (no fallback warning), and the regression test pins both the agreement and the absolute +522.5 s offset for the CoRoT-2 geometry so a sign error can't return. Full measurement table and cause chain in #1390. Test suites: only the two pre-existing `*windows*` failures, unrelated (platform-specific, run on Linux).

### DIFF
--- a/exotic/exotic.py
+++ b/exotic/exotic.py
@@ -21399,6 +21399,20 @@ def process_flat_frames(flat_files, master_bias=None):
 def convert_jd_to_bjd(non_bjd, p_dict, info_dict):
     global _BJD_FALLBACK_WARNING_LOGGED
 
+    # Coerce sexagesimal-string coordinates to numeric degrees before any
+    # conversion. If a caller path delivers RA as "HH:MM:SS.ss" (the -ov -nea
+    # flag combination did exactly this, see issue), barycorrpy raises a
+    # TypeError and the astropy fallback below would silently parse the RA
+    # string as DEGREES (19:27:06 -> 19.45 deg instead of 291.78 deg),
+    # computing the barycentric correction for a point far away on the sky
+    # and shifting BJD_TDB by hundreds of seconds.
+    p_dict = dict(p_dict)
+    if isinstance(p_dict.get('ra'), str) or isinstance(p_dict.get('dec'), str):
+        p_dict['ra'], p_dict['dec'] = radec_hours_to_degree(
+            p_dict.get('ra'), p_dict.get('dec'), non_interactive_run=True,
+            target_name=p_dict.get('pName'),
+        )
+
     try:
         goodTimes = JDUTC_to_BJDTDB(non_bjd, ra=p_dict['ra'], dec=p_dict['dec'], lat=info_dict['lat'],
                                     longi=info_dict['long'], alt=info_dict['elev'])[0]
@@ -31076,7 +31090,11 @@ def _main_impl():
             CandidatePlanetBool = False
 
         if file_cmd_opt == 2:
-            if args.nasaexoarch:
+            # -nea combined with -ov: the override wins the data choice above
+            # (pDict = userpDict, sexagesimal strings), so the coordinate
+            # conversion must still run or string RA/Dec reach the time
+            # conversion and airmass code unconverted.
+            if args.nasaexoarch and not args.override:
                 pass
             elif args.override:
                 try:
--- a/tests/test_radec_non_interactive.py
+++ b/tests/test_radec_non_interactive.py
@@ -73,3 +73,28 @@ def test_invalid_target_coordinates_abort_when_archive_coordinates_unavailable(m
             non_interactive_run=True,
             target_name='Example b',
         )
+
+
+def test_convert_jd_to_bjd_coerces_sexagesimal_strings():
+    """Regression: with -ov -nea the main flow used to deliver sexagesimal
+    RA/Dec STRINGS to convert_jd_to_bjd. barycorrpy then raised TypeError and
+    the astropy fallback parsed "19:27:06.50" as 19.45 DEGREES instead of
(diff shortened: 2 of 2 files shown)


opusbuilds:
Re-verified against the current tip (d295d8c, 4.4.0), per the always-test-latest rule: the bug is still present there. Repro on the tip: string coordinates crash barycorrpy (TypeError), the astropy fallback engages and mis-parses hours as degrees — `"19:27:06.50"` vs its numeric equivalent differ by **−198.7 s** of BJD_TDB for the CoRoT-2 geometry. The branch cherry-picks onto d295d8c cleanly (no conflicts; verified locally, acceptance test passes: string-vs-numeric agree to 0.000 s). Happy to rebase the PR itself onto the tip if that's preferred.

opusbuilds:
Re-verified on the current tip `9100cd3` just now (my previous comment cited `d295d8c`, which was two days stale — my error, corrected). The bug is unchanged there: string coordinates crash barycorrpy with `TypeError: can't multiply sequence by non-int of type 'float'`, the astropy fallback engages, and `"19:27:06.50"` vs its numeric equivalent still differ by **−198.7 s** of BJD_TDB for the CoRoT-2 geometry.

The branch cherry-picks onto `9100cd3` cleanly (no conflicts), and the acceptance test passes on the resulting build (`4.3.2.dev120`): string and numeric agree to 0.000 s. Happy to rebase the PR onto the tip if you'd rather review it that way.

PR #1402, opened 2026-08-19
VSP auto-added comparison stars must sit 5% clear of frame edges

opusbuilds:
As requested in-channel ("comp stars perhaps need to not be within 5% of the edge of the frame... do a pull request... Old leaf head that is") — here it is.

**Motivating case:** Amadeus's 2026-08-16 CoRoT-2 run. After the VSX check correctly removed a variable comp, the VSP fallback auto-added `[99, 1]` — one pixel from the frame edge — on a night whose field drifts ~73 px in y (measured between my solved first/last frames of the same directory). The comp exits the frame under drift and the reduction errors after the star was already accepted. Old filter: `1 < pixel < axis`.

**Change:** candidates must clear every edge by `VSP_COMPARISON_EDGE_MARGIN_FRACTION = 0.05` of that dimension (named constant; ≥32 px in x, ≥25 px in y on a 650×500 MObs frame). Implemented as a small helper with garbage-input safety, applied at the existing in-frame filter — one call site, no behavior change for interior candidates.

**Tests:** `tests/test_vsp_edge_margin.py` pins the motivating coordinates, the exact margin boundary (accepted at the line, rejected inside it), all four edges independently, and non-numeric/None inputs returning False rather than raising. 7 passed against the current tip (`fcd35c8`; branch is cut from it as of this hour).

One judgment call to flag: the margin applies to VSP **auto-added** candidates only — user-supplied comps are untouched, since a human placing a comp near an edge may know their frames don't drift. If you'd rather warn on those too, happy to extend.

### DIFF
--- a/exotic/exotic.py
+++ b/exotic/exotic.py
@@ -17881,8 +17881,7 @@ def vsp_query(file, axis, obs_filter, img_scale, maglimit=14, user_comp_stars=No
             ra_pixel = float(np.asarray(ra_pix, dtype=float).reshape(-1)[0])
             dec_pixel = float(np.asarray(dec_pix, dtype=float).reshape(-1)[0])
             if not (
-                1 < ra_pixel < axis[0]
-                and 1 < dec_pixel < axis[1]
+                vsp_candidate_clear_of_edges(ra_pixel, dec_pixel, axis)
                 and obs_filter in [band['band'] for band in star['bands']]
             ):
                 continue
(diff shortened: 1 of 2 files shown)


opusbuilds:
Scope note from the in-channel question: the edge margin covers **both** VSP entry points — the opt-in path (`aavso_comp == 'y'`, exotic.py:33064) and the automatic V-calibration fallback (exotic.py:18056), which reaches `vsp_query` with no flag check for V-family observations when NextAstro returns no usable V calibration (documented in the shipped inits guide: "even when 'Add Comparison Stars from AAVSO?' is n"). The motivating [99,1] case arrived through the fallback on default settings, so the fix is live for default-config MObs users, not only for those who opt in.

PR #1407, opened 2026-08-30
LM boundary scout: physical envelope + prior-inconsistency fallback on QC FAIL (#1406)

### LINKED ISSUE #1406: 6efeab8 full mode: LM boundary scout expands a/R* to [1e-6, 18], turning an honest prior-fallback into a confident wrong depth (real-night benchmark, 15.7 min vs 24 min)
## Summary

Benchmarking 6efeab8 ("quick mode") on a real night: the new full mode is **1.5× faster than last week's tip and 3.3× faster than dev121** — but on this marginal night it also **changes the depth answer from an honest prior-fallback to a confidently wrong measurement**. The mechanism is visible in the log: the new LM boundary scout expands the a/R\* search bound to effectively unbounded, and UltraNest then settles into the depth/geometry degeneracy. Quick mode itself looks good and is labeled honestly.

## Setup

Same CoRoT-2 b MicroObservatory night (2026-08-15 UT, 76 frames), byte-identical inits, same machine, as the earlier ultranest/numpy A/Bs on #1403. Three runs compared:

| run | build | wall time |
|---|---|---|
| baseline | dev12x (cee6309), full mode | 24 min |
| full mode | dev130 (6efeab8) | **944 s (15.7 min)** |
| quick mode | dev130, `"quick_look_mode": true` | **72 s** |

## Results

| quantity | baseline (dev12x) | dev130 full | dev130 quick (LM) |
|---|---|---|---|
| QC / KTMF | MARGINAL 3.40 | MARGINAL 3.60 | MARGINAL 3.84 |
| Tmid | 2461268.7914 ± 0.0058 | 2461268.7971 ± 0.0062 | 2461268.7983 ± 0.0049 |
| Rp/R\* | 0.167 ± 0.015 (prior fallback) | **0.274 ± 0.045** | 0.183 ± 0.044 |
| a/R\* | 6.09 ± 0.43 | **5.24 ± 0.27** | — |
| inclination | 84.0 ± 2.3 | **79.4 ± 5.5** | — |
| ΔBIC (transit vs flat) | 8.20 | 4.52 | — |
| residual scatter | 1.82 % | 1.89 % | — |

Literature Rp/R\* for CoRoT-2 b is 0.164–0.167. The dev130 full-mode value is 2.4σ from that by its own bar (a 7.5 % depth vs the real 2.8 %), with a/R\* and inclination dragged along the degeneracy. Tmid moved +8 min, inside the bars: **the timing product is robust; the depth product is not.**

## Mechanism, from the run log

Baseline (dev12x): the Rp/R\* posterior pinned against a bound → retries → `rerunning UltraNest with Rp/R* fixed to the input prior` — i.e. the pipeline *declined to claim a depth* it couldn't measure, and said so (the #1386 behavior working as intended).

dev130 full mode: before UltraNest, the new scout runs and *only expands*:
```
LM boundary scout iteration 1 expanded rprs [0.110, 0.223] -> [0.096, 0.305]; ars [6.03, 7.37] -> [4.69, 8.71]
LM boundary scout iteration 2 expanded rprs [...] -> [0.090, 0.305]; ars [4.69, 8.71] -> [0.67, 12.73]
LM boundary scout iteration 3 expanded ars [0.67, 12.73] -> [1e-06, 18.27]
```
With a/R\* allowed down to 1e-6, the sampler finds a deep-and-grazing solution that fits 1.8 % scatter as well as the physical one, and it never hits a bound, so the prior fallback never fires. The speedup is real (the expanded, well-sized space converges faster), but on a night where the data cannot constrain geometry, "expand until the LM fit stops touching the edge" removes the very signal (edge-pinning) that used to tell the pipeline the depth was unmeasurable.

## Suggestions

1. Cap the scout's expansion physically — a/R\* below ~2 and Rp/R\* above ~0.3 are not plausible for any transiting-planet prior EXOTIC will be handed; `[1e-6, 18.27]` should never be a search range.
2. Keep an edge-pinning-style diagnostic alive even when bounds are expanded: e.g. if the posterior's Rp/R\* differs from the prior by more than N× the prior's own uncertainty *and* the data-derived uncertainty is large, flag "depth poorly constrained" rather than reporting the excursion with a tight bar. This night is a clean test case.
3. Quick mode is well done — LM, clearly labeled preliminary / not submission-ready / not a posterior, Tmid within 1.3σ of the full fits, 72 s. Nothing to change there; it's the right tool for the Colab first-look problem.

Happy to run any variant of this night as a test (e.g. scout with physical caps) — inits and frames are on hand.

opusbuilds:
Fixes #1406. Opened as a draft while the regression pair finishes; numbers will be added and the draft flipped to ready.

## What was wrong

On a marginal night the LM boundary scout expanded a/R\* to `[1e-6, 18.3]` in three iterations, after which UltraNest settled into the deep-grazing degeneracy (Rp/R\* 0.274 ± 0.045 against 0.167 literature) and, never touching a bound, skipped the prior fallback that the previous build had correctly fired on the same frames. Two causes, both floors:

1. The generic per-key expansion passed the global `ARS_SEARCH_BOUND_MIN = 1e-6` as the a/R\* floor. Within three width-sized steps that walks a/R\* to inside the star.
2. The geometry envelope's `physical_lower` fell back to the same `1e-6` whenever its inputs (Rp/R\*, eccentric scale) were unavailable.

A correction to my own issue text: the duration constrains a/R\* only from **above** (for a fixed transit length a smaller a/R\* is always available by raising b and Rp/R\* — that *is* the degeneracy), so the existing central-duration envelope is right where it is and this PR does not touch the ceiling logic. The physical envelope is what excludes the branch.

## What changes

- `lm_boundary_scout_physical_envelope(prior)`: a/R\* floor `max(1.5, 1 + Rp/R*, 0.35 × prior a/R*)`, prior-relative ceiling `3 × prior a/R*` for the generic expansion (duration-derived ceilings may still exceed it); Rp/R\* ceiling `min(RPRS_SEARCH_BOUND_MAX, max(0.30, 2 × prior))`. Absolute floors when no prior exists.
- Generic expansion uses the envelope instead of the global floors.
- Geometry envelope: no `1e-6` fallback; the a/R\* interval is clamped into the envelope unconditionally at the end (`physical_envelope_clamp` adjustment reason, only when it changes something).
- Prior-inconsistency fallback (item 3, in the maintainer's FAIL-only form): if the final Rp/R\* posterior sits more than 3σ (combined with the prior's own uncertainty, from `fit_transit_qc_expected_context`) from the input prior **and** `evaluate_transit_detection_qc(fit)['status'] == 'fail'`, the depth goes through the existing data-only fallback path with edge `'prior-inconsistent'`. MARGINAL nights are exempt.

## Tests

- New `tests/test_lm_boundary_scout_caps.py` (6): envelope values with/without prior, `1 + Rp/R*` never violated, generic expansion clamps to the floor across repeated iterations, geometry envelope never falls back to `1e-6`, ceiling respects the prior multiple.
- `test_lm_boundary_scout_geometry_envelope_covers_central_duration_limit`: the a/R\* floor expectation moves from `1 + Rp/R* = 1.13` to the physical floor (4.26 for that fixture's prior of 12.17). The ceiling assertion (2× central limit) is unchanged.
- Full suite locally: **1068 passed, 0 failed** (with `tests/test_quick_look.py` excluded — it reads stdin during collection under a local `pytest` run; CI is green on it, so that's a local-invocation quirk, noted only).

## Regression pair (running)

Same target, same instrument, one week apart, opposite correct behaviors:
- **2026-08-16 night (MARGINAL):** must not report 0.274; expected prior fallback or flag. On this build the scout's third iteration now reads `ars [2.58, 12.73] -> [2.58, 18.27]` where before it read `[1e-6, 18.27]`.
- **2026-08-09 night (QC PASS):** must keep measuring Rp/R\* ≈ 0.173 freely.

Numbers to follow in a comment.

### DIFF
--- a/exotic/exotic.py
+++ b/exotic/exotic.py
@@ -332,6 +332,21 @@ ULTRANEST_LM_BOUNDARY_SCOUT_KEYS = ('rprs', 'ars', 'tmid')
 ULTRANEST_LM_BOUNDARY_SCOUT_RPRS_PRIOR_LOWER_SCALE = 0.5
 ULTRANEST_LM_BOUNDARY_SCOUT_RPRS_PRIOR_UPPER_SCALE = 1.5
 ULTRANEST_LM_BOUNDARY_SCOUT_ARS_CENTRAL_DURATION_UPPER_SCALE = 2.0
+# Physical envelope for the scout (issue #1406). The global search floors are
+# unphysical for a/R* (ARS_SEARCH_BOUND_MIN = 1e-6), so a scout that expands
+# "within the existing limits" can walk a/R* to the inside of the star and hand
+# UltraNest the deep-grazing degeneracy. These caps are relative to the input
+# prior where one exists, with absolute floors where it does not.
+ULTRANEST_LM_BOUNDARY_SCOUT_ARS_PHYSICAL_FLOOR = 1.5
+ULTRANEST_LM_BOUNDARY_SCOUT_ARS_PRIOR_LOWER_SCALE = 0.35
+ULTRANEST_LM_BOUNDARY_SCOUT_ARS_PRIOR_UPPER_SCALE = 3.0
+ULTRANEST_LM_BOUNDARY_SCOUT_RPRS_ABSOLUTE_CEILING = 0.30
+ULTRANEST_LM_BOUNDARY_SCOUT_RPRS_PRIOR_CEILING_SCALE = 2.0
+# Prior-inconsistency fallback (issue #1406, item 3): when the final Rp/R*
+# posterior sits far from the input prior AND the transit QC is a clear FAIL,
+# route the depth through the same data-only path the edge-pinning fallback
+# uses. MARGINAL nights are left alone on purpose (maintainer's call).
+RPRS_PRIOR_INCONSISTENCY_FALLBACK_SIGMA = 3.0
 FAST_ULTRANEST_BEFORE_FINAL_RUN_DEFAULT = True
 FAST_ULTRANEST_MAX_BINNED_POINTS = 20
 FAST_ULTRANEST_MIN_POINTS_TO_BIN = 60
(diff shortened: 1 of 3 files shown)


opusbuilds:
## Regression pair results (this branch, dev131 + these changes)

| | 08-09 night, QC **PASS** | 08-16 night, QC **MARGINAL** |
|---|---|---|
| expectation | keep measuring ~0.173 freely | must not report 0.274 with a tight bar |
| Rp/R\* | **0.177 ± 0.025** (was 0.173) ✓ free measurement preserved | **0.247 ± 0.045** (was 0.274 on 6efeab8; 0.167 by fallback on cee6309) |
| a/R\* | 6.83 ± 0.47 | 5.23 ± 0.26 |
| inclination | 89.8 ± 3.3 | 79.7 ± 4.5 |
| KTMF | 4.24 PASS | 3.76 MARGINAL |
| scout iteration 3, a/R\* | — | `[2.58, 12.73] → [2.58, 18.27]` (was `[1e-6, 18.27]`) |

**What the pair shows, honestly.** The envelope does what it was written to do: no more `1e-6`, and the PASS night is untouched. But on the MARGINAL night the sampler still settles on the deep-grazing branch — at a/R\* 5.2, which is *inside* any defensible physical envelope. So the physical caps are necessary but not sufficient for this night, and the two triggers both miss it, for reasons worth stating:

- The prior-inconsistency trigger is FAIL-only by design, and this night is MARGINAL (3.76).
- Even if it applied, the Rp/R\* test would not fire: 0.247 ± 0.045 is only **1.8σ** from the prior by its own bar (the widened bar hides the excursion).
- What *does* scream is a/R\*: **5.23 ± 0.26 against a prior of 6.70 is 5.7σ.** The degeneracy moves Rp/R\* and a/R\* together, and a/R\* carries the tighter bar.

**Proposed follow-up (one more commit on this branch if wanted):** extend the inconsistency test to a/R\* as well as Rp/R\* — fire when *either* departs the prior by > 3σ combined — and leave the QC gate to the maintainer: FAIL-only keeps this night's 0.247 as-is; FAIL-or-MARGINAL would route it through the data-only path. The old build's behavior on this exact night (fallback to 0.167 with a data-only bar) is the honest answer for a night whose ΔBIC is 4.3; I'd argue for MARGINAL inclusion when a/R\* is the parameter that's off, but that's a judgment call that belongs upstream.

Flipping this out of draft for the envelope change, which is correct and tested regardless of that decision.

PR #1413, opened 2026-09-09
Initialization-file pre-flight: exotic -pf inits.json

opusbuilds:
Adds `exotic -pf inits.json` (`--preflight`): checks an initialization file before any photometry is run, and exits 1 on a hard failure so a scripted run can gate on it:

```
exotic -pf inits.json && exotic -red inits.json -ov
```

Follows the #exotic thread of 09-08 (Michael: "if its useful in exotic, lets bung it in"; solver order nova first, NextAstro as the fallback).

## What it checks

1. **Archive agreement.** Period (1e-4 relative), Rp/Rs and a/Rs (15%), inclination (2%) against `NASAExoplanetArchive`. Under `-ov` nothing currently compares these, which is how a template copy's ephemeris (the shipped HAT-P-32 b sample's `2.1500082`) survives into a reduction. `check_parameters` only runs without `-ov` and is an interactive prompt. Candidates skip this with a note.
2. **Transit inside the window.** Start/end from every frame's header via `img_time_jd`; epoch nearest the window centre; duration from the existing `estimate_transit_duration_from_prior_geometry`. Fails if the transit does not overlap the window at all; a partial transit is reported for a look, not failed, since the pipeline fits partials.
3. **Seed pixel on the target.** Projects the archive RA/Dec (inits RA/Dec as fallback) through a WCS and compares with `Target Star X & Y Pixel`, tolerance 8 px. Header WCS is used as-is. Otherwise the pipeline's own `get_wcs` runs (nova from a source list first, NextAstro fallback, reversed under `--use-nextastro-astrometry`); since #1414 neither path uploads the image, so the `Plate Solution? (y/n)` answer no longer matters here and the frame is solved exactly as the reduction will solve it. Tries the first three frames by time and says if the seed frame was not frame 1 (EXOTIC seeds from frame 1). Skips cleanly with a note if nothing solves; no local astrometry.net.
4. **Comparison stars on the first frame.** On the frame and below the overexposure threshold (header saturation, e.g. Cecilia 4096, else `saturation_value` at `overexposure_threshold_fraction`, default 0.9) are rules. Brightness relative to the target (5x noise, 0.1-20x) is reported for a look, not failed: one frame is a small sample and on a clouded night the first frame can be the worst one. #1409 rejects off-frame comps after the fact; this says so before the photometry is run.

## Where the code is

- `exotic/api/preflight.py`: the checks as pure functions over plain values (dicts, arrays, a WCS) plus a `PreflightReport` (pass / FAIL / look / skip). No `exotic.exotic` import, so it tests without frames or network.
- `exotic/exotic.py`: `run_inits_preflight` and `_preflight_wcs_for_frame` next to `get_wcs`, the `-pf` argument, and a dispatch at the top of `_main_impl` (guarded with `getattr`, like `quick_look`, for the tests that drive `main()` with a namespace).
- `tests/test_preflight.py`: 27 tests, including three end-to-end runs on synthetic MObs-shaped frames with the archive and the solver monkeypatched (clean file exits 0; template period exits 1 with both the archive and window failures; seed 42 px off plus a comp off the frame exits 1 with both named).

## Evidence

Full suite on this branch: **1179 passed** (tip `0cc46cc` alone: 1152 passed with the same `--ignore=tests/test_quick_look.py`, which needs tkinter this box lacks). The 27 new tests account for the whole difference.

Live run on a real MicroObservatory night (WASP-11 b, 2026-09-05, 93 frames, run before #1414, `Plate Solution? n`, so NextAstro source-list only; ~2 min end to end, most of it the archive scrape):

```
  [ok  ] frames with a readable observation time  93 of 93
  [ok  ] planet found in the NASA Exoplanet Archive  WASP-11 b
  [ok  ] Orbital Period (days)  inits 3.72247975  archive 3.72247975  (0.00% off, tolerance 0.01%)
  [ok  ] Ratio of Planet to Stellar Radius (Rp/Rs)  inits 0.1413  archive 0.14128910290606278  (0.01% off, tolerance 15%)
  [ok  ] Ratio of Distance to Stellar Radius (a/Rs)  inits 12.3  archive 12.3  (0.00% off, tolerance 15%)
  [ok  ] Orbital Inclination (deg)  inits 89.1  archive 89.1  (0.00% off, tolerance 2%)
  [    ] observing window  2461288.82141 -> 2461289.01528  (4.65 h, epoch 1170)
  [ok  ] predicted transit overlaps the window  Tmid 2461288.91662 (+2.29 h from start), duration 2.56 h
  [ok  ] full transit inside the window  ingress +1.00 h  egress +3.57 h from start
  [ok  ] target pixel on the target (frame 1, NextAstro source-list solution)  WCS puts the target at (206.5, 341.4); inits pixel (208, 342) is 1.6 px away (tolerance 8)
  [ok  ] seed frame is the first frame  first frame solved
  [    ] first-frame photometry context  background 426, noise 8, target peak +322 ADU above background
  [ok  ] target unsaturated  peak 748 vs threshold 3686
  [ok  ] comp (388, 471) unsaturated  peak 1024 vs threshold 3686
  [ok  ] comp (388, 471) brightness  +598 ADU above background, 1.9x the target
  ... (six comps, all ok)
All pre-flight checks passed.        exit 0
```

Same night with four errors planted (sample period left in, seed moved 42 px, a comp at [700, 100] on a 650x500 frame):

```
  [FAIL] Orbital Period (days)  inits 2.1500082  archive 3.72247975  (42.24% off, tolerance 0.01%)
  [FAIL] predicted transit overlaps the window  Tmid 2461289.53193 (+17.05 h from start), duration 1.48 h
  [FAIL] target pixel on the target (frame 1, NextAstro source-list solution)  WCS puts the target at (206.5, 341.4); inits pixel (215, 300) is 42.2 px away (tolerance 8)
  [look] comp (388, 471) brightness  +598 ADU above background, 37.4x the target
  [FAIL] comp (700, 100) inside the frame  outside the frame
4 pre-flight check(s) failed. Do not reduce with this file yet.        exit 1
```

The "37x the target" looks are the seed sitting on blank sky (+16 ADU), the same symptom read from the other side.

## One design point to overrule if wanted

- **Partial transits and comp brightness are looks, not failures.** Both are judgment calls in practice; the exit code is reserved for things that are wrong on any reading.

Not touched: the reduction path. `-pf` returns before `reduction_opt` is chosen, so `-red`/`-pre`/`-ql` behave as before.

### DIFF
--- a/exotic/api/preflight.py
+++ b/exotic/api/preflight.py
@@ -1,0 +1,268 @@
+# ########################################################################### #
+#    Copyright (c) 2019-2020, California Institute of Technology.
+#    All rights reserved.  Based on Government Sponsored Research under
+#    contracts NNN12AA01C, NAS7-1407 and/or NAS7-03001.
+#
+#    Redistribution and use in source and binary forms, with or without
+#    modification, are permitted provided that the following conditions
+#    are met:
+#      1. Redistributions of source code must retain the above copyright
+#         notice, this list of conditions and the following disclaimer.
+#      2. Redistributions in binary form must reproduce the above copyright
+#         notice, this list of conditions and the following disclaimer in
+#         the documentation and/or other materials provided with the
+#         distribution.
+#      3. Neither the name of the California Institute of
+#         Technology (Caltech), its operating division the Jet Propulsion
+#         Laboratory (JPL), the National Aeronautics and Space
+#         Administration (NASA), nor the names of its contributors may be
+#         used to endorse or promote products derived from this software
+#         without specific prior written permission.
+#
+#    THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+#    "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+#    LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+#    A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE CALIFORNIA
+#    INSTITUTE OF TECHNOLOGY BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+#    SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
+#    TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+#    PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+#    LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+#    NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+#    SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
+# ########################################################################### #
+#    EXOplanet Transit Interpretation Code (EXOTIC)
+#    # NOTE: See companion file version.py for version info.
+# ########################################################################### #
+"""Pre-flight checks for an initialization file, run before any photometry.
+
+Every check here is cheap (FITS headers, one image, one archive query, at most
+one plate solve) and targets a mistake that -ov otherwise lets through until an
+hour of reduction has been spent on it:
+
+  * planetary parameters copied from a template and never edited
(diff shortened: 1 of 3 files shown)


opusbuilds:
Follow-up queued: once #1414 (nova source-list solver) lands, `_preflight_wcs_for_frame` will just call `get_wcs` and the `Plate Solution?` branch and the design note above go away, since the answer no longer changes what leaves the machine. Holding that commit until #1414 merges so a 'n' user does not upload an image if this lands first.

opusbuilds:
Rebased on WBoM at d1dbb2a (#1414 merged) and added the promised follow-up commit: `_preflight_wcs_for_frame` now just calls `get_wcs`, since neither solver path uploads the image any more. The `Plate Solution? n` design point is gone from the description; 161 tests across pre-flight and plate-solution pass.

PR #1414, opened 2026-09-10
Plate-solve at nova.astrometry.net from a source list, not the image

opusbuilds:
Follows the #1413 thread in #exotic (Michael's "sounds fine", 09-09). `NovaSourceListPlateSolution` builds the same DAOStarFinder list EXOTIC already sends to NextAstro, writes it as a FITS binary table (X, Y, FLUX) and uploads that to nova's `/upload` with `image_width`/`image_height` plus the usual scale and RA/Dec hints. nova recognises an xylist by itself; no other API change. Login, submission and job polling are inherited unchanged from `PlateSolution`, so a slow nova queue hands off to NextAstro at exactly the point an image upload would. The list is ~60x smaller than a frame and no pixels leave the machine.

`get_wcs` now uses it wherever it used the image upload (nova-first default and the NextAstro-first `--use-nextastro-astrometry` fallback). The image class stays in the module, unused by `get_wcs`.

**What I found while factoring the upload payload, and fixed in the same change.** nova reads `allow_commercial_use`, `allow_modifications` and `publicly_visible` from *inside* `request-json` ([net/api.py, `upload`](https://github.com/dstndstn/astrometry.net/blob/main/net/api.py): `json.get('publicly_visible', 'y')`). The old `_upload` sent them as bare form fields beside `request-json`, which nova ignores, so every image EXOTIC has uploaded to nova defaulted to public. They are inside `request-json` now, for both classes; there is a test pinning that.

**Pixel convention.** EXOTIC's extractor is 0-based; nova reads xylists 1-based (FITS). The writer adds 1. On WASP-11 09-05 frame 1 the 1-based list put the target 0.6 px from its seed, the 0-based list 1.6 px. NextAstro's own WCS on this frame matched the 0-based nova result; the payload says `"pixel_indexing": "0-based"`, so either NextAstro reads that field or it is ~1 px off. Worth a look by whoever owns that server; not touched here.

**No cancel.** Checked nova's routes rather than memory (`net/urls.py`, `net/api.py`): login, upload, url_upload, submissions, jobs, calibration, tags, info, annotations, myjobs, jobs_by_tag; web side only has hide/unhide. A submission we stop polling for still runs when its turn comes. With a source list that is about a second of solver time. The docstring says so and names `_sub_status` running out as the place a cancel call would go.

**Live run** (this branch, `NovaSourceListPlateSolution(...).plate_solution()` on WASP-11 09-05 frame 1, 650x500, tonight's queue): solved in 186 s, xylist 11,520 bytes against a 656,640-byte frame, `wcs.fits` written with the image data (500, 650), target at (207.5, 342.4) against seed (208, 342), 0.6 px. Last night's two hand-rolled list uploads took 56 s and 325 s; the wait is nova's queue, not the solve.

**Tests.** `tests/test_nova_source_list.py`: shared extractor gives the identical list to NextAstro; xylist is 1-based with the right columns, image dims and private flags in `request-json`; the image class now puts the flags in `request-json` too; the full inherited login/upload/submission/job/wcs flow against mocked `requests`; a blank frame fails at `Source extraction` before any request. Three `get_wcs` tests in `test_nextastro_variability.py` repointed at the new class. 1154 pass; `test_quick_look.py` needs tkinter, which this venv lacks, and errors on the clean branch too.

**Answered in #exotic (Michael, 09-10): dropped.** The size argument is his and it is the stronger one: the source list scales with the 200-source cap, not the sensor, so it stays ~10 KB while a QHY600 frame is ~250 MB. MicroObservatory's small frames hide the issue; many users' connections would not carry the image at all.

~~**Question for you.** With this in, `get_wcs` never uploads an image. Do you want the image upload kept as an explicit last tier (source list to nova, then NextAstro, then image to nova), or dropped from the path as here? I lean dropped: the failure modes of a source list that does not solve are mostly the same frames an image would not solve, and it keeps the "no pixels leave the machine" statement simple.~~

Once this lands, #1413 gets a follow-up commit: `_preflight_wcs_for_frame` just calls `get_wcs` and the `Plate Solution?` branch goes, since the answer no longer changes what leaves the machine. I am holding that until this merges so a 'n' user does not upload an image if #1413 lands first.

### DIFF
--- a/exotic/api/plate_solution.py
+++ b/exotic/api/plate_solution.py
@@ -35,7 +35,7 @@
 #    EXOplanet Transit Interpretation Code (EXOTIC)
 #    # NOTE: See companion file version.py for version info.
 # ########################################################################### #
-from astropy.io.fits import Header, PrimaryHDU, getdata, getheader
+from astropy.io.fits import BinTableHDU, Column, Header, PrimaryHDU, getdata, getheader
 from astropy.stats import sigma_clipped_stats
 from json import dumps
 from pathlib import Path
@@ -63,7 +63,9 @@ _R_MAX_STOPS_LOW = 7
 _R_MAX_STOPS = 10
 _R_MAX_SECS = 37
 _RQ_TIMEOUT = 16.0
-_NEXTASTRO_MAX_SOURCES = 200
+_MAX_SOURCES = 200
+_NEXTASTRO_MAX_SOURCES = _MAX_SOURCES
+_NOVA_XYLIST_FILENAME = "nova_sources.xyls"
 _NEXTASTRO_STATUS_MAX_POLLS = 60
 _NEXTASTRO_STATUS_POLL_SEC = 2
 _NEXTASTRO_IN_PROGRESS_STATUSES = {'queued', 'running'}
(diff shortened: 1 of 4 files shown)


opusbuilds:
The public-upload finding is filed on its own as #1415, so it survives whatever happens to this PR.

PR #1416, opened 2026-09-12
Carry the pre-fit transit coverage verdict into the final plot title and the AAVSO #NOTES

opusbuilds:
## What

When the pre-UltraNest coverage assessment finds that the observed timestamps do not cover the ephemeris-predicted transit window and does not expect the fit to succeed, put that one line where the observer will see it:

- as a second line of the final light-curve title (`FinalLightCurve_*.png`, and the `_HighRes` copy);
- appended to `#NOTES=` in the AAVSO file (or alone when the observer's notes are empty / `na`).

`transit_coverage_caption(fit)` in `output_files.py` reads the assessment that `annotate_pre_ultranest_transit_coverage` already attaches to the fit. It returns `None` unless the assessment ran, is valid, and `expected_successful` is false, so covered nights produce byte-identical outputs. `#QC-XC` already carries `status` machine-readably; this is the human-readable copy.

Example line:

```
COVERAGE 0%: pre-transit baseline only; fit success VERY LOW
```

(two title lines on the plot, split at the `; `; one line in `#NOTES`, prefixed `EXOTIC`)

## Why (a night from yesterday)

MObs WASP-10 b, 2026-09-11 UT, 29 frames 06:06-07:31 UT. By the archive ephemeris (P 3.0927616, T0 2454664.037295, epoch 2144) ingress is 08:55 and mid-transit 10:02 UT, so the last frame is 84 min before ingress. The log at d77c750 says so, three times, once per comparison candidate:

```
Expected transit window: ingress 2461294.87140196, mid 2461294.91816540, egress 2461294.96492884 BJD_TDB (duration 134.7 min).
Observed coverage: pre-transit baseline only; 0.0% of the expected transit window with 27 pre-ingress, 0 in-transit, and 0 post-egress point(s).
Estimated fit success: VERY LOW (~5%). The observed timestamps do not overlap the expected transit window; UltraNest is unlikely to recover a constrained transit solution.
```

Then transit QC fails on every candidate (KTMF 2.33/5.00) and the run continues "with the best available fit so final outputs are still produced". The outputs: a plot titled `WASP-10 b` with a model ingress drawn through the last five points, `Mid-Transit Time [BJD_TDB]: 2461294.8448 +/- 0.0083` (08:20 UT, 50 min after the last frame, 101 min before the ephemeris, with a 12-minute bar), and an AAVSO file whose `#NOTES` is the observer's own text. Nothing the observer reads says the fit is an extrapolation. The same night, reduced by another observer with the same seed pixel and the same comp, was uploaded to #data-upload-aavso yesterday with `Tmid 2461294.8486 +/- 0.0086`; the two runs agree with each other because both are the same extrapolation.

With this change the plot title carries `COVERAGE 0%: pre-transit baseline only` / `fit success VERY LOW` under the target name, and `#NOTES` ends with `| EXOTIC COVERAGE 0%: pre-transit baseline only; fit success VERY LOW`. (A second run on the same frames with this branch is in progress; I will attach the annotated plot in a comment when it finishes.)

Not in this PR: putting the transit-QC verdict itself on the plot. That may be worth doing too, but it is a separate decision about the KTMF outputs; this one only surfaces a verdict EXOTIC has already made before the fit began.

## Tests

`tests/test_plots.py`: title carries the line for an uncovered fit; unchanged for a covered fit and for a fit without the assessment. `tests/test_output_files.py`: `aavso_notes_text` with observer notes, `na`, empty, a covered fit, and the existing `DummyFit`.

```
python -m pytest -q tests/test_plots.py tests/test_output_files.py
76 passed
```

### DIFF
--- a/exotic/output_files.py
+++ b/exotic/output_files.py
@@ -3827,10 +3827,7 @@ def transit_coverage_caption(fit):
         observed = f"{100.0 * float(assessment.get('transit_fraction_observed')):.0f}%"
     except (TypeError, ValueError):
         observed = "?%"
-    return (
-        f"COVERAGE: {segment}; {observed} of the expected transit window observed; "
-        f"fit success {str(label).upper()}"
-    )
+    return f"COVERAGE {observed}: {segment}; fit success {str(label).upper()}"
 
 
 def aavso_notes_text(notes, fit):
--- a/exotic/output_files.py
+++ b/exotic/output_files.py
@@ -3344,7 +3344,7 @@ class OutputFiles:
                     f"{gaia_pmra_header}"
                     f"{gaia_pmdec_header}"
                     f"#COMP_STAR-XC={dumps(comp_star)}\n"
-                    f"#NOTES={self.i_dict['notes']}\n"
+                    f"#NOTES={aavso_notes_text(self.i_dict['notes'], self.fit)}\n"
                     "#DETREND_PARAMETERS=AIRMASS, AIRMASS CORRECTION FUNCTION\n"  # fixed
                     "#MEASUREMENT_TYPE=Rnflux\n"  # fixed
                     f"#FILTER={self.i_dict['filter']}\n"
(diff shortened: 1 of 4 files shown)


opusbuilds:
The same 29 frames, reduced on this branch (2937ff0):

WASP-10 b 2026-09-11, coverage verdict in the title

`#NOTES` ends with `| EXOTIC COVERAGE 0%: pre-transit baseline only; fit success VERY LOW`.

Three runs of this night today, all with the same inits and the same 0% coverage:

| run | code | Tmid (BJD_TDB) | note |
|---|---|---|---|
| 1 | d77c750 | 2461294.8448 ± 0.0083 | 08:20 UT, 50 min after the last frame |
| 2 | this branch, first caption | 2461294.75 ± 0.47 | the honest bar, by chance of the sampler |
| 3 | this branch, 2937ff0 | 2461294.8450 ± 0.0091 | as run 1 |

Runs 1 and 3 are the answer the observer uploaded yesterday (2461294.8486 ± 0.0086). Run 2 is what the posterior looks like when the local-fit replacement does not fire. Whichever one a given run produces, the title now says why it should not be read as a timing.

